PR #17, opened 2020-07-31
games-util/legendary: use single python target.

ZeroPointEnergy:
@tastytea This is a proposed change to the  games-util/legendary ebuild.

This uses only a single implementation for python for an installation, which makes more sense in my opinion for a enduser program.

- https://dev.gentoo.org/~mgorny/python-guide/eclass.html#single-impl-vs-multi-impl
- https://dev.gentoo.org/~mgorny/python-guide/distutils.html#python-single-r1-variant

### DIFF
--- a/games-util/legendary/legendary-0.0.19-r1.ebuild
+++ b/games-util/legendary/legendary-0.0.19-r1.ebuild
@@ -3,7 +3,8 @@
 
 EAPI=7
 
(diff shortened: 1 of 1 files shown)


tastytea:
Thanks!

ZeroPointEnergy:
yw :-)

PR #21, opened 2020-09-26
Update systemd-swap package, new maintainer

vilgotf:
I'm the upstream maintainer and noticed that the systemd-swap package is severely outdated in gentoo so I fixed it up to the last bash version.

Currently there is a python rewrite but one of its dependencies aren't in gentoo but that's no problem since I'll maintain the bash branch for the time forward.

First time contributing an ebuild change so please tell me if I should reorder the commits or any other best practices :)


AndrewAmmerlaan:
This looks good :+1: 

The guide for committing to ::guru can be found here: https://wiki.gentoo.org/wiki/Project:GURU/Information_for_Contributors

vilgotf:
I think I'm done :+1:

vowstar:
> I think I'm done

You should push to ``git@git.gentoo.org:repo/proj/guru.git`` instead of pull request to github directly, 

https://wiki.gentoo.org/wiki/Project:GURU/Information_for_Contributors

The guide said:

> In order to commit to GURU, you need to have a SSH key (used to access the server) and OpenPGP key (used to sign commits and push certificates). If you don't have OpenPGP key yet, following GLEP 63 rules is recommended. Generating GLEP-63-based OpenPGP keys is a decent guide to creating the key.

First step you should generate an OpenPGP keys.

> Please also make sure to read Certificate of Origin, as you will be required to confirm the acceptance of its terms in every commit. Its purpose is to confirm that your commits do not violate copyright.
Once you're ready, please file an access request, including your real name, preferred e-mail address (you will receive bug mail and other comms on this address) and SSH key. For bonus points, you can also include the fingerprint of your OpenPGP key but this is currently not necessary.

And then [read Certificate of Origin](https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin), [file an access request](https://bugs.gentoo.org/enter_bug.cgi?product=GURU&component=Access+requests), then you could get the access to ``git@git.gentoo.org:repo/proj/guru.git`` , the gentoo-bot will mirror it to github.

PR #31, opened 2021-01-04
profiles: add ability to mask per profile/feature

telans:
I think this requires review.
https://trofi.github.io/posts/201-masking-a-package-in-gentoo-overlay.html

Package-Manager: Portage-3.0.12, Repoman-3.0.2
Signed-off-by: Theo Anderson <telans@posteo.de>


AndrewAmmerlaan:
Wouldn't this affect packages outside of ::guru as well? We shouldn't change the behaviour of ::gentoo packages in ::guru.

telans:
> Wouldn't this affect packages outside of ::guru as well?

How so?

profiles/profiles.desc
```diff
@@ -0,0 +1,5 @@
+amd64   default/linux/amd64/17.1/hardened/selinux               stable
+amd64   default/linux/amd64/17.1/no-multilib/hardened/selinux   stable
+amd64   default/linux/amd64/17.1/selinux                        stable
+x86     default/linux/x86/17.0/hardened/selinux                 stable
+x86     default/linux/x86/17.0/selinux                          stable
```
telans, line 5:
```suggestion
amd64   default/linux/amd64/17.1/hardened/selinux               stable
amd64   default/linux/amd64/17.1/no-multilib/hardened/selinux   stable
amd64   default/linux/amd64/17.1/selinux                        stable
x86     default/linux/x86/17.0/hardened/selinux                 stable
x86     default/linux/x86/17.0/selinux                          stable
```

Perhaps these should be `testing` or something, not sure

AndrewAmmerlaan:
> > Wouldn't this affect packages outside of ::guru as well?
> 
> How so?

Because it changes the profiles, or does that only affect packages in ::guru?

telans:
From what I can tell this only affects ::guru. The `parent` files seem to append items to profiles, but it's just when this repository is enabled. It doesn't modify or conflict with anything from ::gentoo by the looks of it.

I'd like review from a dev or someone who knows about this for clarity though

AndrewAmmerlaan:
> I think a review from a dev or someone who knows about this would be good if it were to be merged

@juippis @mgorny @epsilon-0 any thoughts on this?

epsilon-0:
Sorry, I'm busy with schoolwork for a few weeks, won't be able to give a lot of help. :cry:

metadata/layout.conf
```diff
@@ -11,3 +11,7 @@ sign-manifests = false
 
 # Let's start with new EAPIs if possible
 eapis-banned = 0 1 2 3 4 5
+
+# Part of ability to mask features in overlays
+# https://trofi.github.io/posts/201-masking-a-package-in-gentoo-overlay.html
+profile-formats = portage-2
```
AndrewAmmerlaan, line 17: I can't find any documentation about this on the wiki: https://wiki.gentoo.org/wiki/Repository_format/metadata/layout.conf

Are you sure this is required? Considering that we already have sys-apps/portage-3.X.Y in the tree, this feature might already be integrated into portage by default. And thus this might be unnecessary.

AndrewAmmerlaan: Also, pkgcheck says `UnusedProfileDirs: unused profile dirs: 'features', 'features/selinux'`, so I wonder if this is doing anything at all.

telans: In regards to `UnusedProfileDirs`, `features/selinux/package.mask` must be doing something as without it I get the following:

```
  dependency.bad [fatal]        5
   sys-block/systemd-swap/systemd-swap-4.4.0.ebuild: RDEPEND: ~amd64(default/linux/amd64/17.1/hardened/selinux)
[     'dev-python/python-systemd[python_targets_python3_8(-),-python_single_target_python3_8(-)]']
   sys-block/systemd-swap/systemd-swap-4.4.0.ebuild: RDEPEND: ~amd64(default/linux/amd64/17.1/no-multilib/hardened/selinux)
[     'dev-python/python-systemd[python_targets_python3_8(-),-python_single_target_python3_8(-)]']
   sys-block/systemd-swap/systemd-swap-4.4.0.ebuild: RDEPEND: ~amd64(default/linux/amd64/17.1/selinux)
[     'dev-python/python-systemd[python_targets_python3_8(-),-python_single_target_python3_8(-)]']
   sys-block/systemd-swap/systemd-swap-4.4.0.ebuild: RDEPEND: ~x86(default/linux/x86/17.0/hardened/selinux)
[     'dev-python/python-systemd[python_targets_python3_8(-),-python_single_target_python3_8(-)]']
   sys-block/systemd-swap/systemd-swap-4.4.0.ebuild: RDEPEND: ~x86(default/linux/x86/17.0/selinux)
[     'dev-python/python-systemd[python_targets_python3_8(-),-python_single_target_python3_8(-)]']
```

I would say that the pkgcheck warning is displaying as overlays don't support feature masking without this hacky stuff.

---

Without `profile-formats = portage-2` I get the following issues from repoman:

```
!!! Unable to parse profile: '/home/telans/src/guru/profiles/default/linux/amd64/17.1/hardened/selinux'
!!! ParseError: Parent '/home/telans/src/guru/profiles/default/linux/amd64/17.1/hardened/selinux/gentoo:default/linux/amd64/17.1/hardened/selinux' not found: '/home/telans/src/guru/profiles/default/linux/amd64/17.1/hardened/selinux/parent'
!!! Unable to parse profile: '/home/telans/src/guru/profiles/default/linux/amd64/17.1/no-multilib/hardened/selinux'
!!! ParseError: Parent '/home/telans/src/guru/profiles/default/linux/amd64/17.1/no-multilib/hardened/selinux/gentoo:default/linux/amd64/17.1/no-multilib/hardened/selinux' not found: '/home/telans/src/guru/profiles/default/linux/amd64/17.1/no-multilib/hardened/selinux/parent'
!!! Unable to parse profile: '/home/telans/src/guru/profiles/default/linux/amd64/17.1/selinux'
!!! ParseError: Parent '/home/telans/src/guru/profiles/default/linux/amd64/17.1/selinux/gentoo:default/linux/amd64/17.1/selinux' not found: '/home/telans/src/guru/profiles/default/linux/amd64/17.1/selinux/parent'
!!! Unable to parse profile: '/home/telans/src/guru/profiles/default/linux/x86/17.0/hardened/selinux'
!!! ParseError: Parent '/home/telans/src/guru/profiles/default/linux/x86/17.0/hardened/selinux/gentoo:default/linux/x86/17.0/hardened/selinux' not found: '/home/telans/src/guru/profiles/default/linux/x86/17.0/hardened/selinux/parent'
!!! Unable to parse profile: '/home/telans/src/guru/profiles/default/linux/x86/17.0/selinux'
!!! ParseError: Parent '/home/telans/src/guru/profiles/default/linux/x86/17.0/selinux/gentoo:default/linux/x86/17.0/selinux' not found: '/home/telans/src/guru/profiles/default/linux/x86/17.0/selinux/parent'
```

telans:
The changes here are somewhat covered in the wiki: https://wiki.gentoo.org/wiki/Profile_(Portage)#Combining_profiles

AndrewAmmerlaan:
> The changes here are somewhat covered in the wiki: https://wiki.gentoo.org/wiki/Profile_(Portage)#Combining_profiles

Thanks, I think I sort of understand what is going on here now. I found some more information on the `profile-formats` setting in the portage manual: https://dev.gentoo.org/~zmedico/portage/doc/man/portage.5.html

```
Control functionality available to profiles in this repo such as which files may be dirs, or the syntax available in parent files. Use "portage-2" if you're unsure. The default is "portage-1-compat" mode which is meant to be compatible with old profiles, but is not allowed to be opted into directly. Setting profile-bashrcs will enable the per-profile bashrc mechanism package.bashrc. Setting profile-set enables support for using the profile packages file to add atoms to the @profile package set. See the profile packages section for more information. Setting profile-default-eapi enables support for the profile_eapi_when_unspecified attribute. Setting build-id allows dependency atoms in the profile to refer to specific builds (see the binpkg-multi-instance FEATURES setting in make.conf(5)). A build-id atom is identical to a version-specific atom, except that the version is followed by a hyphen and an integer build-id. 
```

Considering that ::guru is a new repo I don't think we have to worry about "compatibility with old profiles", so we should be fine setting it to `portage-2`. 

Does this new profile show up in `eselect profile list`? If it doesn't I think we can merge this.

telans:
> Does this new profile show up in `eselect profile list`?

```
  [35]  guru:default/linux/amd64/17.1/hardened/selinux (stable)
  [36]  guru:default/linux/amd64/17.1/no-multilib/hardened/selinux (stable)
  [37]  guru:default/linux/amd64/17.1/selinux (stable)
```

I don't suppose there's a way to block this either

Alessandro-Barbieri:
You have not fixed the problem, you have created another. Now the user have to use this guru profile and combine it with his existing profiles

telans:
No, the package is masked regardless of whether that profile is set or not . It just seems that the extra profiles are a byproduct of combining feature masks from multiple overlays. If a way to remove those are found, this will work fine.

AndrewAmmerlaan:
> > You have not fixed the problem, you have created another. Now the user have to use this guru profile and combine it with his existing profiles
> 
> No, the package is masked regardless of whether that profile is set or not. It just seems that the extra profiles are a byproduct of combining feature masks from multiple overlays. If a way to remove those is found, this will work fine.

Maybe we can set a property 'hidden' or something for the profile?

Would something like this also work for:

```
   dependency.badindev           3
   www-apps/code-server-bin/code-server-bin-3.4.1.ebuild: RDEPEND: ~amd64(default/linux/amd64/17.0/x32)
['>=net-libs/nodejs-12.16.1:0/12[ssl]']
   www-apps/code-server-bin/code-server-bin-3.6.1.ebuild: RDEPEND: ~amd64(default/linux/amd64/17.0/x32)
['>=net-libs/nodejs-12.16.1:0/12[ssl]']
   www-apps/code-server-bin/code-server-bin-3.8.0.ebuild: RDEPEND: ~amd64(default/linux/amd64/17.0/x32)
['>=net-libs/nodejs-12.16.1:0/12[ssl]']
```

Since nodejs is masked on the `default/linux/amd64/17.0/x32` profiles?

telans:
> Would something like this also work for:

I can't get it to work. Maybe beause it's not a 'feature'?

https://dpaste.com/BKVJ4XW7Z

It doesn't work when placing `package.mask` into `profiles/default/linux/amd64/17.0/x32` either (with or without the parent files)

https://dpaste.com/9NALPVVSV

telans:
> Would something like this also work for:

It does work: https://dpaste.com/9UEA6H6PF

AndrewAmmerlaan:
> > Would something like this also work for:
> 
> It does work: https://dpaste.com/9UEA6H6PF (but also an extra profile)

Hmm, so we either have to accept that there will be useless profiles shown in `eselect profile list` or we have to accept that some packages depend on a masked package in some profile.

I'm not sure about this one, I'm not that fond of these "ghost profiles" in `eselect profile list`, on the other hand it doesn't seem to hurt anything.

telans:
The extra profiles aren't wanted, but I think it's better than leaving dependency issues in the repo

AndrewAmmerlaan:
From https://wiki.gentoo.org/wiki/Profile_(Portage):

```
The eselect profile list command only shows profiles in profiles.desc files of all repositories configured in /etc/portage/repos.conf, with an architecture field that matches the machine's architecture. These are the ones that in most contexts are referred to as 'profiles' with no further qualification. Profiles not named in profiles.desc might be named in other profiles' parent files, and can be thought of as 'subprofiles' or 'profile building blocks'. 
```

Does this still work without the `profiles.desc` file? Because removing that file seems to be the solution to our "ghost-profile" problem.

telans:
> Does this still work without the `profiles.desc` file? Because removing that file seems to be the solution to our "ghost-profile" problem.

Nope, it goes right back to the original issue for whatever reason.

```
  dependency.bad [fatal]        5
   sys-block/systemd-swap/systemd-swap-4.4.0.ebuild: RDEPEND: ~amd64(default/linux/amd64/17.1/hardened/selinux)
[     'dev-python/python-systemd[python_targets_python3_8(-),-python_single_target_python3_8(-)]']
   sys-block/systemd-swap/systemd-swap-4.4.0.ebuild: RDEPEND: ~amd64(default/linux/amd64/17.1/no-multilib/hardened/selinux)
[     'dev-python/python-systemd[python_targets_python3_8(-),-python_single_target_python3_8(-)]']
   sys-block/systemd-swap/systemd-swap-4.4.0.ebuild: RDEPEND: ~amd64(default/linux/amd64/17.1/selinux)
[     'dev-python/python-systemd[python_targets_python3_8(-),-python_single_target_python3_8(-)]']
   sys-block/systemd-swap/systemd-swap-4.4.0.ebuild: RDEPEND: ~x86(default/linux/x86/17.0/hardened/selinux)
[     'dev-python/python-systemd[python_targets_python3_8(-),-python_single_target_python3_8(-)]']
   sys-block/systemd-swap/systemd-swap-4.4.0.ebuild: RDEPEND: ~x86(default/linux/x86/17.0/selinux)
[     'dev-python/python-systemd[python_targets_python3_8(-),-python_single_target_python3_8(-)]']
```

AndrewAmmerlaan:
What about creating a new arch named `hidden` which does absolutely nothing, and have the `profiles.desc` file like this:

```

hidden   default/linux/amd64/17.1/hardened/selinux                    dev
hidden   default/linux/amd64/17.1/no-multilib/hardened/selinux   dev
hidden   default/linux/amd64/17.1/selinux                                    dev
hidden   default/linux/x86/17.0/hardened/selinux                         dev
hidden   default/linux/x86/17.0/selinux                                         dev
```

If that works our problem is solved as well, because `eselect profiles list` only shows profiles for enabled arches (and I think we can safely assume that no one will set `KEYWORDS="hidden"` in make.conf. It's a bit of a hacky workaround though.

telans:
`repoman: invalid arch: "hidden" in /home/telans/src/guru/profiles/profiles.desc line 1` :(

AndrewAmmerlaan:
> `repoman: invalid arch: "hidden" in /home/telans/src/guru/profiles/profiles.desc line 1`. That would've been a good solution

You'll have to add `hidden` to `profiles/arch.list` and `profiles/arches.desc` I think.

telans:
Only `arch.list` it seems. But it works! No more extra profiles and repoman is happy

telans:
I'll push the fix for it and mask nodejs using this method too.

AndrewAmmerlaan:
> Only `arch.list` it seems. But it works! No more extra profiles and repoman is happy

Adding
```
hidden      testing
```
to `arches.desc` is probably a good idea anyway, because that way we explicitly mark this arch as `testing` and not `stable`

> I'll push the fix for it and mask nodejs using this method too.

Awesome!

telans:
If we don't keyword it, does it default to `stable` or would it be esentially unusable? I think we want it to be impossible to use right?

AndrewAmmerlaan:
> If we don't keyword it, does it default to `stable` or would it be esentially unusable? I think we want it to be impossible to use right?

Stable is the default according to https://www.gentoo.org/glep/glep-0072.html

telans:
rebased, fingers crossed.

profiles/profiles.desc
```diff
@@ -0,0 +1,6 @@
+hidden  default/linux/amd64/17.1/selinux                        stable
+hidden  default/linux/amd64/17.1/hardened/selinux               stable
+hidden  default/linux/amd64/17.1/no-multilib/hardened/selinux   stable
+hidden  default/linux/amd64/17.0/x32                            stable
+hidden  default/linux/x86/17.0/selinux                          stable
+hidden  default/linux/x86/17.0/hardened/selinux                 stable
```
AndrewAmmerlaan, line 6: Could these be `dev` instead of `stable`? Or doesn't that work?

telans: Looks like it can be dev. I was trying to match the parent, but I might change it to `exp` instead?

AndrewAmmerlaan: `exp` would be even better (I was unaware that `exp` was a thing it could be)

telans: Actually doing that might exclude it from `repoman -dx` checks. I'll change it to dev.

AndrewAmmerlaan:
Awesome, this looks good now :+1: 

One more small thing, I think we should add some comments to `arches.desc` and `profiles.desc`. Because while this works, it is also pretty complex, and it will be difficult for anyone to figure out what we did here and why we did it without any comments.

telans:
How does this read?

AndrewAmmerlaan:
> How does this read?

Perfect, this should make clear what we did and why :+1:

Alessandro-Barbieri:
@mgorny what do you think?

mgorny:
I don't have a strong opinion. This `hidden` hack looks fragile. The whole thing relies on Portage-specific stuff but I suppose other PMs will just ignore it, so no real change.

telans:
If there are no objections, I'll push this to dev in a bit

Alessandro-Barbieri:
@mgorny what's a more robust way of doing the same thing that will work with other package managers?

mgorny:
The usual way is to add masks straight into ::gentoo profiles.  Ofc, it somewhat sucks because this triggers non-existing package check.

AndrewAmmerlaan:
@telans I tried to do this exact same thing in ::science, but got this:
```

FAILURE RESULTS
/opt/hostedtoolcache/Python/3.9.2/x64/bin/pkgcheck replay --color y /home/runner/failures.json
sci-biology/ssaha2-bin
  UnstatedIuse: version 2.5.5: attr(fetchables): profile 'default/linux/amd64/17.0/x32' (1 total): unstated flags: [ amd64, x86 ]

sci-biology/trf-bin
  UnstatedIuse: version 4.09: attr(fetchables): profile 'default/linux/amd64/17.0/x32' (1 total): unstated flags: [ amd64, x86 ]

sci-chemistry/chimera-bin
  UnstatedIuse: version 1.14: attr(depend): profile 'default/linux/amd64/17.0/x32' (1 total): unstated flag: [ prefix ]
  UnstatedIuse: version 1.15: attr(depend): profile 'default/linux/amd64/17.0/x32' (1 total): unstated flag: [ prefix ]

sci-chemistry/concoord
  UnstatedIuse: version 2.1: attr(fetchables): profile 'default/linux/amd64/17.0/x32' (1 total): unstated flags: [ amd64, x86 ]
  UnstatedIuse: version 2.1.2: attr(fetchables): profile 'default/linux/amd64/17.0/x32' (1 total): unstated flags: [ amd64, x86 ]

sci-chemistry/nmrpipe
  UnstatedIuse: version 8.7.20200827: attr(rdepend): profile 'default/linux/amd64/17.0/x32' (1 total): unstated flag: [ prefix ]
```
https://github.com/gentoo/sci/commit/3f7fad2f7170cd44f1019865604c6ed196f5b03c

Any idea what is going on?

Repoman is happy, but pkgcheck returns these fatal errors. Looks like this is somehow messing with arch specific use flags, or it might just be a bug in pkgcheck/pkgcore.

telans:
@AndrewAmmerlaan Apologies, I only just got around to checking. I had a look and played with some settings but I have no idea why sorry.

AndrewAmmerlaan:
> @AndrewAmmerlaan Apologies, I only just got around to checking. I had a look and played with some settings but I have no idea why sorry.

I think it's just a bug in pkgcheck, I reported it here: https://github.com/pkgcore/pkgcheck/issues/305

AndrewAmmerlaan:
For some reason this is now happening in ::guru as well: https://github.com/gentoo/guru/runs/2212142180?check_suite_focus=true

AndrewAmmerlaan:
@telans I think we need to re-think this, we seem to have broken something in repoman as well.

So as an example, first `media-video/vidify` was unmasked. It depends on PyQtWebEngine, which is masked on `default/linux/amd64/17.0/x32`. However, this issue did **not** appear in repoman. Because of the issues with pkgcheck I mentioned before, i tried to remove the per-profile masking we introduced, and _then_ the issue with `media-video/vidify` did show up all of a sudden.

I think that instead of actually fixing the issue, we have just broken the profile dependent checking of repoman entirely, which made it seem as if the issue was fixed. This did not show up in pkgcheck because of this bug: https://github.com/pkgcore/pkgcheck/commit/271384f9aff195439442edb4d039432a4eedff7e

PR #38, opened 2021-01-25
nodejs eclass, mocha, pelias and dependencies

Alessandro-Barbieri:
Signed-off-by: Alessandro-Barbieri <lssndrbarbieri@gmail.com>

Comments welcome


eclass/node-guru.eclass
```diff
@@ -0,0 +1,43 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EXPORT_FUNCTIONS src_prepare src_compile src_install
+
+SLOT="0"
```
telans, line 6:
```suggestion
```
should this be set in ebuilds?

eclass/node-guru.eclass
```diff
@@ -0,0 +1,43 @@
...
+IUSE=""
```
telans, line 7:
```suggestion
```
not really needed

eclass/node-guru.eclass
```diff
@@ -0,0 +1,43 @@
...
+DEPEND="net-libs/nodejs"
+RDEPEND="${DEPEND}"
+BDEPEND="
+	app-misc/jq
+	net-misc/rsync
+	sys-apps/moreutils
+"
```
telans, line 14:
```suggestion
NODEJS_DEPEND="net-libs/nodejs"
NODEJS_BDEPEND="
	app-misc/jq
	net-misc/rsync
	sys-apps/moreutils
"
DEPEND="${NODEJS_DEPEND}"
RDEPEND="${NODEJS_DEPEND}"
BDEPEND="${NODEJS_BDEPEND}"
```
maybe something like this so it can be overridden and called from in ebuilds?

eclass/node-guru.eclass
```diff
@@ -0,0 +1,43 @@
...
+}
+
+node-guru_src_install() {
+	#copy files instead of symlinks
+	rsync -avLAX "${T}/prefix/" "${ED}/usr" --exclude /bin || die
+
+	if [ -d "${T}/prefix/bin" ]
+	then
+		#keep the symlinks
+		rsync -avAX "${T}/prefix/bin/" "${ED}/usr/bin" || die
```
telans, line 41: could this be done with `cp -a`?

dev-node/agentkeepalive/agentkeepalive-4.1.3.ebuild
```diff
@@ -0,0 +1,17 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+inherit node-guru
```
telans, line 5:
```suggestion
EAPI=7

inherit node-guru
```

dev-node/agentkeepalive/agentkeepalive-4.1.3.ebuild
```diff
@@ -0,0 +1,17 @@
...
+HOMEPAGE="
+	https://github.com/node-modules/agentkeepalive
+	https://www.npmjs.com/package/agentkeepalive
+"
+SRC_URI="https://registry.npmjs.org/agentkeepalive/-/agentkeepalive-4.1.3.tgz -> ${P}.tgz"
+LICENSE="MIT"
+KEYWORDS="~amd64"
+RDEPEND="
+	${DEPEND}
+"
```
telans, line 17: with the previous depend changes:
```suggestion
SRC_URI="https://registry.npmjs.org/agentkeepalive/-/agentkeepalive-4.1.3.tgz -> ${P}.tgz"

LICENSE="MIT"
SLOT="0"
KEYWORDS="~amd64"
```

telans (commented):
Are you set on the names for the eclass and category? If it were up to me it would be `nodejs.eclass` & `dev-nodejs` (`dev-node` seems alright though)

the eclass needs general documentation too

dev-node/agentkeepalive/agentkeepalive-4.1.3.ebuild
```diff
@@ -0,0 +1,17 @@
...
+
+EAPI=7
+inherit node-guru
+
+DESCRIPTION="Missing keepalive http.Agent"
+HOMEPAGE="
+	https://github.com/node-modules/agentkeepalive
+	https://www.npmjs.com/package/agentkeepalive
+"
+SRC_URI="https://registry.npmjs.org/agentkeepalive/-/agentkeepalive-4.1.3.tgz -> ${P}.tgz"
```
telans, line 12:
```suggestion
SRC_URI="https://registry.npmjs.org/agentkeepalive/-/agentkeepalive-4.1.3.tgz -> ${P}.tar.gz"
```
these could be renamed to match most other distfile file types, but it doesn't really matter

telans, line 12:
```suggestion
SRC_URI="https://registry.npmjs.org/agentkeepalive/-/${P}.tgz -> ${P}.tgz"
```
where possible

dev-node/agentkeepalive/agentkeepalive-4.1.3.ebuild
```diff
@@ -0,0 +1,17 @@
...
+LICENSE="MIT"
+KEYWORDS="~amd64"
```
telans, line 14:
```suggestion
KEYWORDS="~amd64"
RESTRICT="mirror"
```

it might be worth adding this just due to the number of packages being added to guru

dev-node/agentkeepalive/agentkeepalive-4.1.3.ebuild
Alessandro-Barbieri: Ok

eclass/node-guru.eclass
Alessandro-Barbieri: Ok

eclass/node-guru.eclass
Alessandro-Barbieri: I don't plan on making slot aware packages for now

telans:
How are the licenses checked? Or are they consistent across all nodejs packages?

dev-node/agentkeepalive/agentkeepalive-4.1.3.ebuild
Alessandro-Barbieri: Maybe in the eclass

AndrewAmmerlaan:
Could we do something about the `RedundantUriRename` `StaticSrcUri` and `  DESCRIPTION.toolong` warnings?

I know they are just warnings, but IMO it would be nice if we could keep the repoman/pkgcheck output in ::guru to a bare minimum.

Alessandro-Barbieri:
I can change anything.
I generate the ebuilds (once, bumping is done manually) with a python script and then I manually fix the eventual issues.
The python script fetches metadata from https://registry.npmjs.org/ that also contains the package license

eclass/node-guru.eclass
```diff
@@ -0,0 +1,43 @@
...
+	#restore original package.json
+	jq 'with_entries(if .key == "deps" then .key = "dependencies" else . end)' package.json | sponge package.json || die
+}
+
+node-guru_src_install() {
+	#copy files instead of symlinks
+	rsync -avLAX "${T}/prefix/" "${ED}/usr" --exclude /bin || die
+
+	if [ -d "${T}/prefix/bin" ]
+	then
```
telans, line 39:
```suggestion
	if [[ -d ${T}/prefix/bin ]]; then
```

AndrewAmmerlaan:
> The python script fetches metadata from https://registry.npmjs.org/ that also contains the package license

Could you also add the script to the repository? Because that will make it easier for everyone to keep this up to data and/or add more packages.

Alessandro-Barbieri:
`SRC_URI` can be put in the eclass.like
`SRC_URI="https://registry.npmjs.org/${PN}/-/${P}.tgz `
and will work with most of the packages
Packages like `@hapi/joi` ebuild named `hapi+joi` need special treatment and the `->`rename to avoid collision

eclass/node-guru.eclass
Alessandro-Barbieri: Ok

Alessandro-Barbieri:
I would like to read more comments on the technical part and from more persons.

Open issues for now
- category name: `dev-node` vs `dev-nodejs`
- eclass name
- `SRC_URI` in the eclass or in the ebuilds

AndrewAmmerlaan:
>     * eclass name

I agree with @telans , there's no reason why eclasses in ::guru would have to be appended with `-guru`. any ebuild in overlay X will always use the eclass in overlay X before it uses the eclass in one of the overlays defined in the `masters` variable of overlay X's `layout.conf`. Now if there were an eclass with the exact same name in ::gentoo that would be a different story, than it might be a good idea to append `-guru` to avoid confusion, but I don't think that is the case here.



>     * `SRC_URI` in the eclass or in the ebuilds

I would define the default in the eclass, an ebuild can always overwrite this if needed anyway.

Alessandro-Barbieri:
Round 2, major changes done

eclass/node-guru.eclass
Alessandro-Barbieri: See the new changes

eclass/node.eclass
```diff
@@ -0,0 +1,49 @@
...
+
+RESTRICT+=" mirror"
+SLOT="0"
+
+[ -z "${SRC_URI}" ] && SRC_URI="https://registry.npmjs.org/${PN}/-/${P}.tgz"
+
+[ -z "${NODEJS_BDEPEND}" ] && BDEPEND+=" ${NODEJS_BDEPEND}"
+[ -z "${NODEJS_RDEPEND}" ] && RDEPEND+=" ${NODEJS_RDEPEND}"
+[ -z "${NODEJS_DEPEND}" ] && DEPEND+=" ${NODEJS_DEPEND}"
+[ -z "${RDEPEND}" ] && DEPEND+=" ${RDEPEND}"
```
AndrewAmmerlaan, line 14: This is probably not necessary, if the variable is set in the ebuild it overwrites the variable set in the eclass anyway, so I think you can just set it unconditionally in the eclass.

dev-node/anymatch/anymatch-3.1.1.ebuild
```diff
@@ -0,0 +1,18 @@
...
+HOMEPAGE="
+	https://github.com/micromatch/anymatch
+	https://www.npmjs.com/package/anymatch
+"
+LICENSE="ISC"
+KEYWORDS="~amd64"
+NODEJS_RDEPEND="
+	dev-node/normalize-path
+	dev-node/picomatch
+"
```
telans, line 18: I don't think your script is appending a newline if dependencies are defined

dev-node/assert-plus/assert-plus-1.0.0.ebuild
```diff
@@ -0,0 +1,14 @@
...
+EAPI=7
+
+inherit node
+
+DESCRIPTION="Extra assertions on top of node's assert module"
+HOMEPAGE="
+	https://github.com/mcavage/node-assert-plus
+	https://www.npmjs.com/package/assert-plus
+"
+LICENSE="MIT"
```
telans, line 13:
```suggestion
HOMEPAGE="
	https://github.com/mcavage/node-assert-plus
	https://www.npmjs.com/package/assert-plus
"

LICENSE="MIT"
```
I'd love a space here

eclass/node.eclass
```diff
@@ -0,0 +1,49 @@
...
+	jq 'if .dependencies? then .dependencies[] = "*" else . end' package.json | sponge package.json || die
+	default
+}
+
+node_src_compile() {
+	#here we trick npm into believing there are no dependencies so it will not try to fetch them
+	jq 'with_entries(if .key == "dependencies" then .key = "deps" else . end)' package.json | sponge package.json || die
+
+	export npm_config_prefix="${T}/prefix"
+	npm install -g --production || die
```
telans, line 35:
```suggestion
	npm install --global --production || die
```
for clarity

eclass/node.eclass
```diff
@@ -0,0 +1,49 @@
...
+
+node_src_prepare() {
+	#remove version constraints on dependencies
+	jq 'if .dependencies? then .dependencies[] = "*" else . end' package.json | sponge package.json || die
+	default
+}
+
+node_src_compile() {
+	#here we trick npm into believing there are no dependencies so it will not try to fetch them
+	jq 'with_entries(if .key == "dependencies" then .key = "deps" else . end)' package.json | sponge package.json || die
```
telans, line 32: why remove the version constraints in src_prepare if npm won't see them anyway? couldn't this be achieved with `sed -i 's/dependencies/dep/' package.json` without `app-misc/jq` or `sys-apps/moreutils`? (also should probably move this to src_prepare regardless)

eclass/node.eclass
```diff
@@ -0,0 +1,49 @@
...
+
+	#restore original package.json
+	jq 'with_entries(if .key == "deps" then .key = "dependencies" else . end)' package.json | sponge package.json || die
+}
+
+node_src_install() {
+	#copy files instead of symlinks
+	rsync -avLAX "${T}/prefix/" "${ED}/usr" --exclude /bin || die
+
+	if [ -d "${T}/prefix/bin" ] ; then
```
telans, line 45:
```suggestion
	if [[ -d ${T}/prefix/bin ]] ; then
```
bashisms

eclass/node.eclass
Alessandro-Barbieri: Version constraints are also checked at runtime.
I'm using jq because it's the right tool to manipulate JSON files.

eclass/node.eclass
Alessandro-Barbieri: Who cares? The interpreter will always be bash

telans: It matches every other eclass

eclass/node.eclass
Alessandro-Barbieri: I'm not getting it. What do you propose to write?

dev-node/anymatch/anymatch-3.1.1.ebuild
Alessandro-Barbieri: It's because I deleted code manually and sometimes the newline disappeared. I'll fix it later, now I'm writing stuff while ssh-ing from a smartphone

eclass/node.eclass
Alessandro-Barbieri: Other eclasses are not even consistent https://github.com/gentoo/gentoo/blob/master/eclass/perl-functions.eclass#L270

telans: I reckon it would be if it had been meaningfully edited in the last four years

Alessandro-Barbieri:
As for now the dependencies variables seems to not work and are hiding dependencies from repoman's check

telans (commented):
> As for now the dependencies variables seems to not work and are hiding dependencies from repoman's check

suggestions below should fix that.

Without any dependencies set in the ebuild, the dependencies default to nodejs, jq, etc. If you need extra dependencies then you also need to call ${NODEJS_{,R,B}DEPEND}. This matches how ${PYTHON_DEPEND} works

dev-node/11ty+eleventy/11ty+eleventy-0.11.1.ebuild
```diff
@@ -0,0 +1,77 @@
+# Copyright 1999-2021 Gentoo Authors
```
telans, line 1: is this a good name for a package? could it be 11ty-eleventy?

dev-node/11ty+eleventy/11ty+eleventy-0.11.1.ebuild
```diff
@@ -0,0 +1,77 @@
...
+
+DESCRIPTION="Transform a directory of templates into HTML."
+HOMEPAGE="
+	https://www.11ty.dev/
+	https://www.npmjs.com/package/@11ty/eleventy
+"
+
+first="${PN%%+*}"
+second="${PN#*+}"
+SRC_URI="https://registry.npmjs.org/@${first}/${second}/-/${second}-${PV}.tgz -> ${P}.tgz"
```
telans, line 16:
```suggestion
if [[ ${PN} =~ '+' ]]; then
    PN_FIRST="${PN%%+*}"
    PN_SECOND="${PN#*+}"
    SRC_URI="https://registry.npmjs.org/@${PN_FIRST}/${PN_SECOND}/-/${PN_SECOND}-${PV}.tgz -> ${P}.tgz"
fi
```

could something like that be moved to the eclass instead?

dev-node/11ty+eleventy/11ty+eleventy-0.11.1.ebuild
Alessandro-Barbieri: Original name is `@11ty/eleventy`

AndrewAmmerlaan: > Original name is `@11ty/eleventy`

In that case I would just go with `eleventy`

eclass/node.eclass
AndrewAmmerlaan: > I'm not getting it. What do you propose to write?

I meant the same thing that @telans suggested below. There's no need to check if the variable is set, ebuild variables overwrite eclass variables anyway, so there is no risk of the eclass overwriting anything.

dev-node/11ty+eleventy/11ty+eleventy-0.11.1.ebuild
Alessandro-Barbieri: `eleventy` ia another package
https://www.npmjs.com/package/@11ty/eleventy
https://www.npmjs.com/package/eleventy

eclass/node.eclass
Alessandro-Barbieri: This code is no more

dev-node/assert-plus/assert-plus-1.0.0.ebuild
Alessandro-Barbieri: Done

dev-node/11ty+eleventy/11ty+eleventy-0.11.1.ebuild
```diff
@@ -0,0 +1,77 @@
...
+
+LICENSE="MIT"
+KEYWORDS="~amd64"
+IUSE="test"
+RESTRICT="!test? ( test )"
+
+NODEJS_BDEPEND="
```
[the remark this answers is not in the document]
Alessandro-Barbieri: Thanks

dev-node/11ty+eleventy/11ty+eleventy-0.11.1.ebuild
```diff
@@ -0,0 +1,77 @@
...
+		dev-node/nyc
+		dev-node/prettier
+		dev-node/rimraf
+		dev-node/toml
+		dev-node/viperhtml
+		dev-node/vue
+		dev-node/vue-server-renderer
+	)
+"
+NODEJS_RDEPEND="
```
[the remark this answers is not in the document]
Alessandro-Barbieri: Thanks

eclass/node.eclass
Alessandro-Barbieri: This code is no more

dev-node/assert-plus/assert-plus-1.0.0.ebuild
Alessandro-Barbieri: Done

eclass/node.eclass
```diff
@@ -0,0 +1,49 @@
...
+[ -z "${NODEJS_RDEPEND}" ] && RDEPEND+=" ${NODEJS_RDEPEND}"
+[ -z "${NODEJS_DEPEND}" ] && DEPEND+=" ${NODEJS_DEPEND}"
+[ -z "${RDEPEND}" ] && DEPEND+=" ${RDEPEND}"
+
+DEPEND+=" net-libs/nodejs"
+BDEPEND+="
+	app-misc/jq
+	net-misc/rsync
+	sys-apps/moreutils
+"
```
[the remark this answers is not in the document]
Alessandro-Barbieri: Thanks alot

Alessandro-Barbieri:
Should I continue with the test dependencies or just avoid testing the packages?

Alessandro-Barbieri:
Could you try running tests  for `pelias-openstreetmap`?

AndrewAmmerlaan:
> I tried testing but there is an issue.
> I want to check the useflag `test` in the install phase even for packages not having tests.
> Currently I get this error `USE Flag 'test' not in IUSE for dev-node/lodash-4.17.20`

The test use flag has to be added manually to ebuilds with:
```
IUSE="test"
RESTRICT="!test? ( test )"
```

The restrict part is necessary to make sure that `FEATURES="test"` is only used if `USE="test"` is also set.

AndrewAmmerlaan:
Nodejs has been masked on amd64/x32 profiles, so we should add packages that depend on it to `profiles/default/linux/amd64/17.0/x32/package.mask` using @telans method of profile specific masking in overlays.

telans:
> The test use flag has to be added manually to ebuilds with:

`IUSE="test"` can be added to the eclass along with having a separate `IUSE` within the ebuild. They don't overwrite eachother

telans:
> Nodejs has been masked on amd64/x32 profiles

Hopefully this can be done with `dev-node/*`

Alessandro-Barbieri:
> > Nodejs has been masked on amd64/x32 profiles
> 
> Hopefully this can be done with `dev-node/*`

I've tried and doesn't work

Alessandro-Barbieri:
I want something to toggle the installation of devDependencies but only when tests are needed

AndrewAmmerlaan:
> I want something to toggle the installation of devDependencies but only when tests are needed

```
USE="...test..."
RESTRICT="!test? ( test )"

BDEPEND="test? ( ... )"
```

should do the trick, right?

Alessandro-Barbieri:
> Nodejs has been masked on amd64/x32 profiles, so we should add packages that depend on it to `profiles/default/linux/amd64/17.0/x32/package.mask` using @telans method of profile specific masking in overlays.

I have added packages to it but is it working?

Alessandro-Barbieri:
> > I want something to toggle the installation of devDependencies but only when tests are needed
> 
> ```
> USE="...test..."
> RESTRICT="!test? ( test )"
> 
> BDEPEND="test? ( ... )"
> ```
> 
> should do the trick, right?

You can see what I've done here https://github.com/gentoo/guru/blob/23727f15499adf5e983f528790a2df019468b767/eclass/node.eclass#L43

Alessandro-Barbieri:
I've just made a new overlay
https://github.com/Alessandro-Barbieri/node-overlay

PR #45, opened 2021-04-01
Update repoman.yml

scardracs:
Updated portage to latest available

### DIFF
--- a/.github/workflows/repoman.yml
+++ b/.github/workflows/repoman.yml
@@ -19,7 +19,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install lxml pyyaml
-        wget -qO - "https://github.com/gentoo/portage/archive/portage-3.0.12.tar.gz" | tar xz
+        wget -qO - "https://github.com/gentoo/portage/archive/portage-3.0.18.tar.gz" | tar xz
         sudo groupadd -g 250 portage
         sudo useradd -g portage -d /var/tmp/portage -s /bin/false -u 250 portage
     - name: Setup master gentoo repository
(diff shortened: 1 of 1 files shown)


AndrewAmmerlaan:
Looks good to me :+1: 

I wonder if it would be possible to have this always fetch the latest version, that way we won't have to touch it every time there is a new version of repoman.

(as a side note, if you want to run the GitHub Actions checks against this PR than you'll have to make the PR against the dev branch instead of the master branch)

scardracs:
> Looks good to me +1
> 
> I wonder if it would be possible to have this always fetch the latest version, that way we won't have to touch it every time there is a new version of repoman.
> 
> (as a side note, if you want to run the GitHub Actions checks against this PR than you'll have to make the PR against the dev branch instead of the master branch)

Yeah I know that (I mean about your side note), but this isn't actually a PR for an ebuild, so I thought I have to PR directly inside master

AndrewAmmerlaan:
> Yeah I know that (I mean about your side note), but this isn't actually a PR for an ebuild, so I thought I have to PR directly inside master. I'm sorry if I have make it wrong

Non-ebuild contributions also go through dev first (otherwise the branches would diverge). You can just commit/PR this just as you would with an ebuild :)

scardracs:
> > Yeah I know that (I mean about your side note), but this isn't actually a PR for an ebuild, so I thought I have to PR directly inside master. I'm sorry if I have make it wrong
> 
> Non-ebuild contributions also go through dev first (otherwise the branches would diverge). You can just commit/PR this just as you would with an ebuild :)

Ok then :). Also I may suggest for you first point to set a "default" portage version and run an action to update it directly from gentoo. This will take surely more time, but we are sure to run repoman against latest version available

PR #46, opened 2021-04-01
Update repoman.yml

scardracs:
@AndrewAmmerlaan something like that

### DIFF
--- a/.github/workflows/repoman.yml
+++ b/.github/workflows/repoman.yml
@@ -31,8 +31,8 @@ jobs:
         sudo ln -s /var/db/repos/gentoo/profiles/default/linux/amd64/17.1 /etc/portage/make.profile
     - name: Update portage
       run: |
-        python3 portage-portage-3.0.18/bin/emerge --sync
-        python3 portage-portage-3.0.18/bin/emerge -1 sys-apps/portage    
+        sudo python3 portage-portage-3.0.18/bin/emerge --sync
+        sudo python3 portage-portage-3.0.18/bin/emerge -1 sys-apps/portage    
     - name: Test with repoman
       run: |
         python3 portage-portage-3.0.18/repoman/bin/repoman full -dx
--- a/.github/workflows/repoman.yml
+++ b/.github/workflows/repoman.yml
@@ -31,7 +31,7 @@ jobs:
         sudo ln -s /var/db/repos/gentoo/profiles/default/linux/amd64/17.1 /etc/portage/make.profile
     - name: Update portage
       run: |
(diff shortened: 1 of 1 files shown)


AndrewAmmerlaan:
I think we might be able to do this faster and easier by querying GitHub for which version is the latest and then downloading that. Something like this might work:
https://gist.github.com/steinwaywhw/a4cd19cda655b8249d908261a62687f8
https://github.com/marketplace/actions/fetch-latest-release

Alternatively we could just `git sync` the `gentoo/portage` repository and then use `git tag | grep  repoman- | sed 's/repoman-//g'` to find the latest version and switch to that using e.g. `git checkout tags/repoman-3.0.3 -b repoman-3.0.3`

scardracs:
> I think we might be able to do this faster and easier by querying GitHub for which version is the latest and then downloading that. Something like this might work:
> https://gist.github.com/steinwaywhw/a4cd19cda655b8249d908261a62687f8
> https://github.com/marketplace/actions/fetch-latest-release
> 
> Alternatively we could just `git sync` the `gentoo/portage` repository and then use `git tag | grep repoman- | sed 's/repoman-//g'` to find the latest version and switch to that using e.g. `git checkout tags/repoman-3.0.3 -b repoman-3.0.3`

Using sed seems promising, let me take some checks :)

scardracs:
Looks like these: https://gist.github.com/steinwaywhw/a4cd19cda655b8249d908261a62687f8 doesn't work for mirrored github repos

scardracs:
Using my way requires to change emerge --sync from rsync to git, so it is a no go (too much effort for what we really need)

AndrewAmmerlaan:
Fixed in: https://gitweb.gentoo.org/repo/proj/guru.git/commit/?h=dev&id=292e03dc22ddc1d5a3bf143d311344f8cc0d9f75

The repoman check should now always use the latest version :)

scardracs:
> Fixed in: https://gitweb.gentoo.org/repo/proj/guru.git/commit/?h=dev&id=292e03dc22ddc1d5a3bf143d311344f8cc0d9f75
> 
> The repoman check should now always use the latest version :)

Looks promising. Have you already tested it?

AndrewAmmerlaan:
> Looks promising. Have you already tested it?

Yes, I have tested this in this PR first: https://github.com/gentoo/guru/pull/49

scardracs:
> > Looks promising. Have you already tested it?
> 
> Yes, I have tested this in this PR first: https://github.com/gentoo/guru/pull/49

Gg!

PR #49, opened 2021-04-02
Update repoman Action

AndrewAmmerlaan:
Lets test this in a PR first before pushing it.

Comments and suggestions are more than welcome :)

### DIFF
--- a/scripts/setup-and-run-repoman.sh
+++ b/scripts/setup-and-run-repoman.sh
@@ -1,0 +1,30 @@
+#! /usr/bin/env bash
+# Maintainer: Andrew Ammerlaan <andrewammerlaan@riseup.net>
+#
+# This sets up repoman and runs the latest version
+#
+# TODO: Force repoman to output in colour
+
+### Setup prerequisites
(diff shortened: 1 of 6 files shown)


scripts/setup-and-run-repoman.sh
```diff
@@ -0,0 +1,33 @@
...
+export TERM=xterm-color
+
+### Setup prerequisites
+python3 -m pip install --upgrade pip
+pip install lxml pyyaml
+sudo groupadd -g 250 portage
+sudo useradd -g portage -d /var/tmp/portage -s /bin/false -u 250 portage
+
+### Sync the portage repository
+git clone https://github.com/gentoo/portage.git
```
tormath1, line 16: what's the base image ? We sure `git` is installed ?

AndrewAmmerlaan: Presumably git is installed since we are using `actions/checkout@v2` as well. In any case that part seems to be working. There are two things which I'm still working on:
- colour output
- figure out why `--jobs=2` makes it slower instead of faster

PR #51, opened 2021-04-06
sys-boot/bmaptools: new package

scardracs:
Bmaptool is a generic tool for creating the block map (bmap) for a file and copying files using the block map. The idea is that large files, like raw system image files, can be copied or flashed a lot faster and more reliably with bmaptool than with traditional tools, like "dd" or "cp".

Package-Manager: Portage-3.0.18, Repoman-3.0.3
Signed-off-by: Marco Scardovi <marco@scardovi.com>


AndrewAmmerlaan:
Looks good to me :+1: 

As a general note on styling, try to stick to the ebuild skeleton as much as possible, it makes it faster to read/debug things when things are where one expects them to be: https://wiki.gentoo.org/wiki/Basic_guide_to_write_Gentoo_Ebuilds#How_to_create_an_ebuild

scardracs:
> Looks good to me :+1: 
> 
> As a general note on styling, try to stick to the ebuild skeleton as much as possible, it makes it faster to read/debug things when things are where one expects them to be: https://wiki.gentoo.org/wiki/Basic_guide_to_write_Gentoo_Ebuilds#How_to_create_an_ebuild

Ok :)

AndrewAmmerlaan:
Ooh just one small thing, if you move/rename a package, you should add a entry to `profiles/updates/...`. That way users who have your package installed will automatically move to the renamed package.

scardracs:
> Ooh just one small thing, if you move/rename a package, you should add a entry to `profiles/updates/...`. That way users who have your package installed will automatically move to the renamed package.

I don't have idea on how to do it

AndrewAmmerlaan:
> > Ooh just one small thing, if you move/rename a package, you should add a entry to `profiles/updates/...`. That way users who have your package installed will automatically move to the renamed package.
> 
> I don't have idea on how to do it

Adding:
```
move app-editors/atom app-editors/atom-bin
```
to `profiles/updates/2Q-2021` should do the trick

app-editors/atom-bin/atom-bin-1.55.0-r1.ebuild
```diff
@@ -0,0 +1,72 @@
...
+src_install(){
+	insinto /opt/"${PN}"
+	doins -r "${S}"/*
+	dosym ../../opt/"${PN}"/atom "${EPREFIX}"/usr/bin/atom
+	fperms +x /opt/"${PN}"/atom
+	dosym ../../opt/"${PN}"/resources/app/apm/bin/apm "${EPREFIX}"/usr/bin/apm
+	fperms +x /opt/"${PN}"/resources/app/apm/bin/apm
+	dosym ../../opt/"${PN}"/resources/app/apm/bin/node "${EPREFIX}"/usr/bin/node
+	fperms +x /opt/"${PN}"/resources/app/apm/bin/node
+	dosym ../../opt/"${PN}"/resources/app/apm/bin/npm "${EPREFIX}"/usr/bin/npm
```
Alessandro-Barbieri, line 68: are you sure this isn't conflicting with nodejs[npm]?

scardracs: I don't know as I don't use nodejs

Alessandro-Barbieri: Is it bundling the while nodejs? How can it run otherwise?

PR #67, opened 2021-06-20
Openstack

Alessandro-Barbieri:


Alessandro-Barbieri:
@prometheanfire what do you think?

prometheanfire:
What is the end goal for this?  the packages still need to be maintained

Alessandro-Barbieri:
Would you like to continue maintaining them here? Or have you completely given up?

prometheanfire:
Gentoo tends to deprecate stable python faster than what upstream openstack targets (they are removing py37 soon iirc).  Upstream tends to stick to what is included in debian and ubuntu.  The python team won't let me package openstack without their oversight.

Alessandro-Barbieri:
@prometheanfire I've restored the ebuilds here. Feel free to make any changes, I'll improve documentation building

prometheanfire:
I think, if you want to keep it around, it is best put into a overlay.  thanks for the interest though.

PR #72, opened 2022-07-01
net-p2p/monero: fix build for musl

alfredfo:
This patch sets -DSTACK_TRACE=OFF when elibc_musl is
defined. Execinfo is not provided by musl libc and therefore the build
fails.

Other, and probably better solutions would be to not depend on
execinfo or to package a standalone
libexecinfo (https://github.com/mikroskeem/libexecinfo) and add it to
the DEPEND-array if using musl. But that's better to take upstream. I
think this is totally fine for most users for now.

Signed-off-by: Alfred Persson Forsberg <cat@catcream.org>

### DIFF
--- a/net-p2p/monero/monero-0.17.3.2.ebuild
+++ b/net-p2p/monero/monero-0.17.3.2.ebuild
@@ -52,6 +52,8 @@ src_configure() {
 		-DUSE_DEVICE_TREZOR=OFF
 	)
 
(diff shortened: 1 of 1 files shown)


alfredfo:
I PR:ed as I usually do with fixes for ::gentoo. Will follow this next time https://wiki.gentoo.org/wiki/Project:GURU/Information_for_Contributors (if you prefer that?)

alfredfo:
Commited to dev branch instead, closing.

PR #74, opened 2022-08-28
sys-power/tlp: Fix systemd support

jayofdoom:
Prior to this change, systemd units were not properly installed.


jayofdoom:
this was pushed to dev, was PR only for review purposes

PR #85, opened 2022-11-04
databases.eclass: new eclass for running databases

CyberTailor:
Here you can see example usage. I will push this myself after review.


eclass/databases.eclass
```diff
@@ -0,0 +1,429 @@
+# Copyright 2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+# @ECLASS: databases.eclass
+# @MAINTAINER:
+# Anna <cyber+gentoo@sysrq.in>
+# @AUTHOR:
+# Anna <cyber+gentoo@sysrq.in>
+# @SUPPORTED_EAPIS: 8
+# @BLURB: eclass to test databases
```
mgorny, line 10: I suppose "test packages against databases" would be more correct.

eclass/databases.eclass
```diff
@@ -0,0 +1,429 @@
...
+			;;
+		get-pidfile)
+			echo "${T}"/run/${funcname}.pid
+			;;
+		get-sockdir)
+			echo "${T}"/run/${funcname}/
+			;;
+		get-sockfile)
+			echo "${T}"/run/${funcname}.sock
+			;;
```
mgorny, line 160: How about putting them all in a single directory, e.g. `"${T}/${funcname}"/...`?

eclass/databases.eclass
```diff
@@ -0,0 +1,429 @@
...
+
+# @FUNCTION: _ememcached_start
+# @USAGE: [port]
+# @INTERNAL
+# @DESCRIPTION:
+# Start memcached server.
+_ememcached_start() {
+	local port=${1:-11211}
+	local pidfile=$(ememcached get-pidfile)
+	mkdir -p "$(dirname "${pidfile}")" || die
```
mgorny, line 185: You could move creating the directory to the common function. I suppose it makes sense to create the parent dir for all databases, even if they (for some reason) do that for you.

eclass/databases.eclass
```diff
@@ -0,0 +1,429 @@
...
+		--daemon
+		--port=${port}
+		--user=nobody
+		--listen=127.0.0.1
+		--pidfile="${pidfile}"
+	)
+
+	ebegin "Spawning memcached"
+	memcached "${myargs[@]}" > /dev/null
+	eend $? || die -n
```
mgorny, line 197: The `die` line here doesn't provide enough context to tell what failed, so you need a specific message.

eclass/databases.eclass
```diff
@@ -0,0 +1,429 @@
...
+		--fork
+	)
+
+	ebegin "Spawning mongodb"
+	mongod "${myargs[@]}" > /dev/null
+	eend $? || die -n
+}
+
+# @FUNCTION: _emongod_stop
+# @USAGE:
```
mgorny, line 240: Remove empty `@USAGE`.

eclass/databases.eclass
```diff
@@ -0,0 +1,429 @@
...
+# ==============================================================================
+
+# @FUNCTION: _epostgresql_start
+# @USAGE: [port]
+# @INTERNAL
+# @DESCRIPTION:
+# Create a new PostgreSQL database and start PostgreSQL server.
+_epostgresql_start() {
+	local port=${1:-5432}
+	local dbpath="$(epostgresql get-dbpath)"
```
mgorny, line 336: Inconsistent quoting.

mgorny (changes requested):
To be honest, I'd love to see this in `::gentoo`. We certainly have a use for it.

eclass/databases.eclass
CyberTailor: Missed in v2, resolved here

eclass/databases.eclass
```diff
@@ -0,0 +1,505 @@
...
+			;;
+		--get-pidfile)
+			echo "${T}"/${funcname}/run/${funcname}.pid
+			;;
+		--get-sockdir)
+			echo "${T}"/${funcname}/run/
+			;;
+		--get-sockfile)
+			echo "${T}"/${funcname}/run/${funcname}.sock
+			;;
```
mgorny, line 233: Still, do you need the `log` and `run` subdirs? It feels like overkill.

CyberTailor: OK, removed `log` and `run` from paths

eclass/databases.eclass
```diff
@@ -0,0 +1,505 @@
...
+#
+# ...
+#
+# inherit databases distutils-r1
+#
+# ...
+#
+# BDEPEND="$(eredis --get-depend)"
+#
+# distutils_enabe_tests pytest
```
mgorny, line 85: Typo.

eclass/databases.eclass
```diff
@@ -0,0 +1,505 @@
...
+	mkdir -p "${dbpath}" || die "Creating database directory failed"
+	mkdir -p "${sockdir}" || die "Creating sockets directory failed"
+
+	# files
+	local logfile_dir=$(dirname "$(${funcname} --get-logfile)")
+	local pidfile_dir=$(dirname "$(${funcname} --get-pidfile)")
+	local sockfile_dir=$(dirname "$(${funcname} --get-sockfile)")
+	mkdir -p "${logfile_dir}" || die "Creating log directory failed"
+	mkdir -p "${pidfile_dir}" || die "Creating pidfile directory filed"
+	mkdir -p "${sockfile_dir}" || die "Creating sockets directory filed"
```
mgorny, line 173: You can collapse these into a single`mkdir` now, since they're all using the same directory, and avoid `dirname` in the process.

CyberTailor: Done

PR #88, opened 2023-01-29
app-misc/ani-cli EOL

medanisjbara:
ani-cli v3.4 and below reached EOL. See their PR [948](https://github.com/pystardust/ani-cli/issues/948)


N-R-K:
Any specific reason for picking 9999 instead of v4.0 ?

medanisjbara:
I didn't notice that they made a release tag for it. Sorry about that.

N-R-K:
No issue. live ebuilds cannot be keyworded so they typically require some hassle from the user's end to install - which is why I pointed it out. If there's any specific problem in v4.0 - you can also make a snapshot ebuild.

medanisjbara:
I just updated the commit with another v4 ebuild which I just tested. It is identical to the other ebuild (since from the beginning it was made with tagged releases in mind).
I'm hoping this is good then.

N-R-K:
Please change the commit message to `app-misc/ani-cli: add 4.0 & 9999, drop old`.

And I believe you will also need to "signoff" your commit (but I'm not too sure if that's still required or not - and the gentoo wiki is down so I cannot check right now).

medanisjbara:
Should be done. Both signoff and the commit message.

app-misc/ani-cli/ani-cli-4.0.ebuild
```diff
@@ -0,0 +1,33 @@
+# Copyright 2022 Gentoo Authors
```
N-R-K, line 1: 2022-2023

app-misc/ani-cli/ani-cli-4.0.ebuild
```diff
@@ -0,0 +1,33 @@
...
+	inherit git-r3
+else
+	SRC_URI="https://github.com/pystardust/ani-cli/archive/refs/tags/v${PV}.tar.gz -> ${P}.tar.gz"
+	KEYWORDS="~amd64 ~x86"
+fi
+LICENSE="GPL-3"
+SLOT="0"
+IUSE=""
+
+COMMUN_DEPEND="
```
N-R-K, line 19: this var seems redundant. I'd inline it to `RDEPEND` directly

PR #90, opened 2023-03-21
fnm: add ppc64 support

darkbasic:
I've patched the fnm ebuild to build on ppc64le, which requires overriding the ring (indirect) dependency with a git branch containing patches. It builds fine if you don't use the cargo eclass, which downloads a tarball from github instead of cloning, leading to the same exact files except for the `.git` directory. Somehow without the `.git` directory it fails to compile... It's not a matter of permissions or anything else: copying the .git directory fixes it.
Log/backtrace: https://pastebin.com/cRGzVsHE

Some help would be appreciated to get rid of the workaround so that I can craft something mergeable.


MrRoy:
@darkbasic is this something you are still interested in getting merged?

darkbasic:
Yes but I have a new patch for this.

MrRoy:
@darkbasic if you update the PR with your new patch we could get this merged, if it's ready

MrRoy:
@darkbasic any update? this PR hasn't had activity for over a year. If you want to merge this, can you update the PR with the new patch and also target a version that's in tree (1.35.1 at the moment)?

darkbasic:
> if you update the PR with your new patch we could get this merged, if it's ready

@MrRoy I've created a new PR to showcase the new approach. We managed to get ppc64 support upstream in ring 0.17, so I've simply bumped the ring version. Not sure if it's an acceptable solution though, but it works.

Link to the new PR: https://github.com/gentoo/guru/pull/188

PR #95, opened 2023-04-27
net-p2p/trezord-go: manifest updated

salfter:
The hashes for the source tarball have changed, so the ebuild won't work until the manifest is updated.


joecool1029:
While this is normally a trivial thing to update alarm bells should be going off like crazy when release hashes change after the fact (as we have no idea if the release is being maliciously tampered with). In this case they probably re-tagged the release but I raised an issue with their upstream and it looks like @CyberTailor has the new hash pending in dev branch. 

Finally, GURU operates differently than the main gentoo repo and doesn't usually check github PR's.  If you'd like to collaborate in the future check out https://wiki.gentoo.org/wiki/Project:GURU

stkw0:
Upstream issue: https://github.com/trezor/trezord-go/issues/293

salfter:
Tried updating the ebuild and found upstream is broken: https://github.com/trezor/trezord-go/issues/297

MrRoy:
This was done under commit 4de6bd98b61574e8c7050ab547788e4532080401
We can close this, I think?

MrRoy:
@salfter can you close this PR please?

PR #99, opened 2023-07-08
games-emulation/duckstation: ebuild updated

kwhat:
This resolves https://bugs.gentoo.org/883721


stkw0:
Thank you for your work. Maybe you may be interested in https://wiki.gentoo.org/wiki/Project:GURU/Information_for_Contributors It's specially important the bit about the Certificate of Origin (https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin)

PR #100, opened 2023-07-10
app-misc/ani-cli: bump to 4.5 && fix non-unique file names in ebuild

medanisjbara:


app-misc/ani-cli/ani-cli-4.5.ebuild
```diff
@@ -10,8 +10,8 @@ if [[ "${PV}" == *9999* ]]; then
 	inherit git-r3
 else
 	SRC_URI="
-		https://github.com/pystardust/${PN}/releases/download/v${PV}/ani-cli
-		https://github.com/pystardust/${PN}/releases/download/v${PV}/ani-cli.1
+		https://github.com/pystardust/${PN}/releases/download/v${PV}/ani-cli -> ani-cli-${PV}
```
N-R-K, line 13: You can use just `${P}` which contains the package name and version. https://devmanual.gentoo.org/ebuild-writing/variables/#predefined-read-only-variables

N-R-K (changes requested):

app-misc/ani-cli/ani-cli-4.5.ebuild
medanisjbara: Thanks. Fixed !

PR #101, opened 2023-07-25
games-emulation/citra: Add dev-util/spirv-headers as a build dependency

MagelessMayhem:
`dev-util/spirv-headers` is required for the ebuild to get to the `src_compile()` phase. It will fail otherwise on current git builds of Citra.


zhuyifei1999:
I commited this to the dev branch. Thanks for the fix!

MagelessMayhem:
No problem!

PR #107, opened 2023-09-05
dev-python/gbinder: Update BDEPEND limit of cython

xz-dev:
gbinder will build fail at cpython3, check https://github.com/erfanoabdi/gbinder-python/pull/14


stkw0:
Seems that upstream is unmaintained, right? I think that for such a small package if it's not even maintained, maybe would make sense to remove altogether it in a near future.

xz-dev:
> Seems that upstream is unmaintained, right? I think that for such a small package if it's not even maintained, maybe would make sense to remove altogether it in a near future.

Maybe we should change the upstream if anyone maintain it, I do not think we should delete it, because it's a part of waydroid.

PR #108, opened 2023-09-11
dev-util/insomnia-bin: add 2023.5.8

martin-kokos:
Last version is about 3 years old


stkw0:
Thank you for your. In case you didn't already, it may be interesting to read https://wiki.gentoo.org/wiki/Project:GURU/Information_for_Contributors It's specially important the bit about the Certificate of Origin (https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin). You can use `pkgdev commit` which among other things will automatically use the proper commit format.

I saw that this version of the ebuild almost don't have any dependencies compared to the previous versions, is that correct?

stkw0:
Seems it doesn't work. At least for me. The previous version seems to work fine though. 
```
[24933:0914/111319.938119:FATAL:spawn_subprocess.cc(221)] posix_spawn: Permission denied (13)
zsh: trace trap  insomnia
```

martin-kokos:
I was going off of ebuild in pentoo overlay (which has a bit more recent version) which wasn't a great idea.
I will recheck it.
However I have no clue what could cause your error? Are you using hardened? I removed pax-utils (because pax doesn't appear to get much use nowdays) which sounds like only thing that could be related.

stkw0:
I thought the same but I tried running "pax-mark -m ...." but the issue was still present. I am using desktop/plasma profile. If you can not reproduce this issue I will try running this ebuild with a clean environment. Which profile are you using?

martin-kokos:
> I thought the same but I tried running "pax-mark -m ...." but the issue was still present. I am using desktop/plasma profile. If you can not reproduce this issue I will try running this ebuild with a clean environment. Which profile are you using?

default/linux/amd64/17.1/desktop/plasma (stable)
(Wayland)

martin-kokos:
I've added dependencies which I found via this command
`objdump -p insomnia | grep NEEDED | awk '{print $2}' | xargs -I % sh -c 'echo %; find /usr -name %| xargs qfile| cut -d":" -f 1; echo' `

martin-kokos:
I have report from another user the ebuild works on a freshly installed desktop profile

stkw0:
I don't understand how it worked for you. Can you check the permissions of /opt/Insomnia/chrome_crashpad_handler?? It was .rw-r--r--  after emerging the package. Insomnia binary ` execve("/opt/Insomnia/chrome_crashpad_handler", ["/opt/Insomnia/chrome_crashpad_ha...` but as that file didn't had execution permissions it reported an error. chmod +x fixed the issue for me

martin-kokos:
I also don't have +x on chrome_crashpad_handler and it works fine by either running `insomnia` or via the .desktop shortcut.
```
.rw-r--r-- root root 1.2 MB Mon Sep 11 18:42:47 2023  /opt/Insomnia/chrome_crashpad_handler
```
the permission denied is also present but not fatal
```
$ insomnia
insomnia: /usr/lib64/libp11-kit.so.0: no version information available (required by /usr/lib64/libgnutls.so.30)
/opt/Insomnia/insomnia: /usr/lib64/libp11-kit.so.0: no version information available (required by /usr/lib64/libgnutls.so.30)
/opt/Insomnia/insomnia: /usr/lib64/libp11-kit.so.0: no version information available (required by /usr/lib64/libgnutls.so.30)
[16986:0919/143332.971072:FATAL:spawn_subprocess.cc(221)] posix_spawn: Permission denied (13)
14:33:32.985 › Running version 2023.5.8
14:33:33.637 › [electron client protocol] FAILED to set default protocol 'insomnia://'
14:33:33.660 › [electron client protocol] the current executable is not the default protocol for 'insomnia://'
14:33:34.210 › [electron client protocol] the default application set for 'insomnia://' is 'insomnia.desktop
'
14:33:34.523 › [fix] Running database repairs
14:33:34.897 › [db] No workspace parentId found for wrk_17039f398f984cc4a0b4d5b1d50e4a86 setting default proj_default-project
14:33:35.100 › [db] Initialized DB at /home/x/.config/Insomnia/insomnia.$TYPE.db
14:33:35.101 › [db] Init responses DB
14:33:35.101 › [db] Init websocket-responses DB
14:33:35.169 › [localstorage] Initialized at /home/x/.config/Insomnia/localStorage
14:33:35.234 › [main] Loading file:///opt/Insomnia/resources/app.asar/index.html
14:33:35.260 › [updater] Not supported on this platform linux
14:33:38.686 › [main] Window ready, handling command line arguments [ '/opt/Insomnia/insomnia' ]
14:33:38.693 › [main] Checking for newer version than  2023.5.8
14:33:40.143 › [main] No newer version

```
Can you try without zsh (although I am also using zsh) to rule that out?

Even if I add +x to chrome_crashpad_handler, the posix_spawn: Permission denied message is gone, but in insomnia the process args for crashpad-handler-pid is 0 either way, so not sure the crash handler would even work in my case.

martin-kokos:
I've updated the ebuild to add x for crash handler.

stkw0:
It happens the same with bash. I think that, although I don't use the hardened profile, I may have turned on some checks which may produce a fatal error on weird conditions like this. In any case, I think it's good to catch this kind of things before merging the PR. Thanks

PR #110, opened 2023-09-27
gui-libs/xdg-desktop-portal-hyprland: Added patch for clang, introduced check for >=clang:17

Fuchskaute:
The build issues with clang have been resolved upstream in the following commits:
https://github.com/hyprwm/xdg-desktop-portal-hyprland/commit/54b5467ec12fb115b537e0c86256226474edf8bb
https://github.com/hyprwm/xdg-desktop-portal-hyprland/commit/f9461b0b7d3071347390feee534638555ff76c7d

I combined the last 3 commits into a patch and added it to the ebuild for 1.1.0 if the user is using clang.
I also added a check for >=clang-17.0.1 as it failed to build with clang-16.0.6 while testing in a chroot.

I would be happy about any feedback or helping with further testing that is required.

### DIFF
--- a/gui-libs/xdg-desktop-portal-hyprland/files/xdg-desktop-portal-hyprland-1.1.0_fix_clang.patch
+++ b/gui-libs/xdg-desktop-portal-hyprland/files/xdg-desktop-portal-hyprland-1.1.0_fix_clang.patch
@@ -1,0 +1,207 @@
+From 54b5467ec12fb115b537e0c86256226474edf8bb Mon Sep 17 00:00:00 2001
+From: Jan Beich <jbeich@FreeBSD.org>
+Date: Mon, 18 Sep 2023 19:00:21 +0000
+Subject: [PATCH 1/3] build: Unbreak build on FreeBSD (#98)
+
+* helpers: add missing header after 9dfb9e125bf0
+
+src/helpers/MiscFunctions.cpp: In function 'void addHyprlandNotification(const std::string&, float, const std::string&, const std::string&)':
+src/helpers/MiscFunctions.cpp:24:9: error: 'fork' was not declared in this scope
+   24 |     if (fork() == 0)
+      |         ^~~~
+src/helpers/MiscFunctions.cpp:25:9: error: 'execl' was not declared in this scope
+   25 |         execl("/bin/sh", "/bin/sh", "-c", CMD.c_str(), nullptr);
+      |         ^~~~~
+
+* build: add missing dependency for libc++ after 2f48e6550374
+
+ld.lld: error: undefined symbol: pthread_create
+>>> referenced by PortalManager.cpp
+>>>               src/xdg-desktop-portal-hyprland.p/core_PortalManager.cpp.o:(CPortalManager::startEventLoop())
+>>> referenced by PortalManager.cpp
+>>>               src/xdg-desktop-portal-hyprland.p/core_PortalManager.cpp.o:(CPortalManager::startEventLoop())
+---
+ CMakeLists.txt                | 4 +++-
+ src/helpers/MiscFunctions.cpp | 1 +
+ src/meson.build               | 1 +
+ 3 files changed, 5 insertions(+), 1 deletion(-)
+
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index e39b72c..44d1689 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -31,12 +31,14 @@ message(STATUS "Checking deps...")
+ add_subdirectory(subprojects/sdbus-cpp)
+ add_subdirectory(hyprland-share-picker)
+ 
++find_package(Threads REQUIRED)
++
+ find_package(PkgConfig REQUIRED)
+ pkg_check_modules(deps REQUIRED IMPORTED_TARGET wayland-client wayland-protocols libpipewire-0.3 libspa-0.2 libdrm gbm)
+ 
+ file(GLOB_RECURSE SRCFILES CONFIGURE_DEPENDS "src/*.cpp")
+ add_executable(xdg-desktop-portal-hyprland ${SRCFILES})
+-target_link_libraries(xdg-desktop-portal-hyprland PRIVATE rt sdbus-c++ PkgConfig::deps)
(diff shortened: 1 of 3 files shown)


stkw0:
I don't understand the reasoning behind having a pkg_setup with those checks instead of having a USE flag like "clang" and having something like 
```
clang? ( >=sys-devel/clang-17.0.1 )
!clang ( >=sys-devel/gcc-13.0.0 )
```

stkw0:
I don't think it's a big issue, but I don't understand the reason, although I believe if someone wrote it that way it is because there is a reason. In general, it's better to fail as soon as possible. Moving those checks to BDEPEND could make life easier for end users as portage will automatically pull the required compiler without the need for the user to run portage, wait while it resolve the dependencies, then have a failure right after starting to emerging packages and then having to manually installing the required compiler and running portage again.

Could you fixup the last commit so the history is cleaner?

Fuchskaute:
If you could define what you mean by fixup gladly! Also I just got GURU commit access myself, so sorry for creating a PR here I got lucky someone actually checked.

micielski:
Commited!

> I don't understand the reasoning behind having a pkg_setup with those checks instead of having a USE flag like "clang" and having something like
> 
> ```
> clang? ( >=sys-devel/clang-17.0.1 )
> !clang ( >=sys-devel/gcc-13.0.0 )
> ```
Well, the thing you are proposing isn't the best (depending on a compiler doesn't mean that the ebuild will compile with it, which is why there is a check in pkg_setup), but you're correct. There should be a dependency for a newer version of gcc/clang, which I've added a few hours ago. Thank you for keeping an eye on me.

Fuchskaute:
Closed by the following commit https://github.com/gentoo/guru/commit/2487eafda90ed51e414083ee0a11e71918cbdb18

Fuchskaute:
Did I make a big mistake

micielski:
> Did I make a big oopsie

Not really, Larry will close this issue anyway, so it doesn't matter whether you leave it open or closed :D

Fuchskaute:
Alright, thank you for the info! I already thought it would be auto-closed but wasn't so I was unsure

stkw0:
> Commited!
> 
> > I don't understand the reasoning behind having a pkg_setup with those checks instead of having a USE flag like "clang" and having something like
> > ```
> > clang? ( >=sys-devel/clang-17.0.1 )
> > !clang ( >=sys-devel/gcc-13.0.0 )
> > ```
> 
> Well, the thing you are proposing isn't the best (depending on a compiler doesn't mean that the ebuild will compile with it, which is why there is a check in pkg_setup), but you're correct. There should be a dependency for a newer version of gcc/clang, which I've added a few hours ago. Thank you for keeping an eye on me.

I don't know if I understood you correctly. The check you are referring can be done by portage directly. In that regard, making the check on pkg_setup don't seem to bring any benefit. I mean, this
```
if [[ $(clang-major-version) -lt 17 ]] ; then
	die "XDPH-1.1.0 requires >= sys-devel/clang-17.0.1 to build"
fi
```
is strictly equivalent to
`BDEPEND=">=sys-devel/clang-17.0.0"` 
afaik, with the only difference that in the second case portage will try to resolve that dependency if possible and if not it will fail when resolving the dependencies instead. Making the check on pkg_setup tricks portage into thinking that the dependencies have been resolved when it may not be the case.

stkw0:
> If you could define what you mean by fixup gladly! Also I just got GURU commit access myself, so sorry for creating a PR here I got lucky someone actually checked.

when making a PR request it may be common that additional small changes are made. Those changes, if small, maybe don't need to have a commit by itself. One common approach is to fixup the commits. I am bad explaining this kind of things but you can search for more info, sry. In practice what I do is to run `git rebase -i HEAD~N` where N is the number of commits and then fixup the commits that have small or trivial changes. The interactive menu already shows a good explanation of the different options.

Fuchskaute:
> > If you could define what you mean by fixup gladly! Also I just got GURU commit access myself, so sorry for creating a PR here I got lucky someone actually checked.
> 
> when making a PR (and only on PR ) request it may happen that additional small changes are made. Those changes, if small, maybe don't need to have a commit by itself. One common approach is to fixup the commits. I am bad explaining this kind of things but you can search for more info, sry. In practice what I do is to run `git rebase -i HEAD~N` where N is the number of commits and then fixup the commits that have small or trivial changes. The interactive menu already shows a good explanation of the different options.

That explanation was already very helpful, thank you!

micielski:
> I don't know if I understood you correctly. The check you are referring can be done by portage directly. In that regard, making the check on pkg_setup don't seem to bring any benefit. I mean, this
> 
> ```
> if [[ $(clang-major-version) -lt 17 ]] ; then
> 	die "XDPH-1.1.0 requires >= sys-devel/clang-17.0.1 to build"
> fi
> ```
> 
> is strictly equivalent to `BDEPEND=">=sys-devel/clang-17.0.0"` afaik, with the only difference that in the second case portage will try to resolve that dependency if possible and if not it will fail when resolving the dependencies instead. Making the check on pkg_setup tricks portage into thinking that the dependencies have been resolved when it may not be the case.

Let's say that I've upgraded Clang from 16 to 17. For a brief moment, I'll have two different versions of that same compiler until I do `emerge --depclean` which will remove the old version, and so there's no guarantee that I'll immediately use the new version of Clang (and I could still overwrite this with yet another compiler in `package.env`, for example with `dev-lang/tcc`).
Here's an example of the same check in [::gentoo](https://github.com/gentoo/gentoo/blob/1cf7e6d4fddfa1868718baf2ac3263782a0b6b74/dev-libs/botan/botan-3.1.1.ebuild#L69C1-L77C4)
Does it make sense? If not, maybe let's hop on IRC? I'm `micielski` on Libera (well, we've already chatted there, but just in case x) )

Fuchskaute:
For further reference here is the same check I introduced in the Hyprland ebuild on ::gentoo.

https://github.com/gentoo/gentoo/blob/1cf7e6d4fddfa1868718baf2ac3263782a0b6b74/gui-wm/hyprland/hyprland-0.29.1.ebuild#L81

stkw0:
I see, that makes sense. Thank you very much for your explanation! :)

PR #112, opened 2023-10-01
updated gui-apps/packettracer to 8.2.1

DaniD3v:
removed 8.2.0 because you can only download 8.2.1 on Ciscos website anyways so there's no point in keeping the old version.


DaniD3v:
bump? why isn't this getting merged?

PR #115, opened 2023-10-23
app-containers/waydroid: Add new runtime dependency

MagelessMayhem:
Fixes a ModuleNotFoundError seen when attempting to initialze Waydroid.

Signed-off-by: Sebastian France <MagelessMayhem@protonmail.com>


MagelessMayhem:
The intent of doing this is to allow Waydroid to find the DBus frontend it needs, which is supplied by `dev-python/dbus-python`. Without the dependency the user is required to go out of their way to install the package themselves.

The ebuild still installs the program normally (which is to be expected) but the program fails to set itself up when running `waydroid init` as root.

PR #116, opened 2023-11-05
dev-util/bruno: New package

martin-kokos:
Bruno is an HTTP API testing GUI tool  similar to Insomnia or Postman.

https://www.usebruno.com/

Do not merge until this issues is resolved: https://github.com/usebruno/bruno/issues/895


martin-kokos:
> Is your email registered in bugzilla (see https://wiki.gentoo.org/wiki/Project:GURU/Information_for_Contributors)? You will not receive bugs regarding this package otherwise.

Yes, its registered. (I've fixed the incorrect email in the CoO)

stkw0:
By the way, in metadata.xml there is a use flag description that is not used in the ebuild.

martin-kokos:
> By the way, in metadata.xml there is a use flag description that is not used in the ebuild.

Ah, so thats where it was. I was looking for int in the ebuild and overlooked it in metadata. Thanks, removed.

martin-kokos:
Bumped to 1.6.1 and included upstream dev as maintainer per request.

dev-util/bruno-bin/metadata.xml
```diff
@@ -0,0 +1,15 @@
...
+        <name>Anoop M D</name>
+    </maintainer>
+	<maintainer type="person">
+		<email>martin-kokos@protonmail.com</email>
+		<name>Martin Mokry</name>
+	</maintainer>
+	<upstream>
+		<remote-id type="github">usebruno/bruno</remote-id>
+	</upstream>
+</pkgmetadata>
```
MrRoy, line 15: Does the upstream dev use Gentoo? If not, you should put them in the <upstream> section instead.
See: https://devmanual.gentoo.org/ebuild-writing/misc-files/metadata/index.html
```suggestion
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE pkgmetadata SYSTEM "https://www.gentoo.org/dtd/metadata.dtd">
<pkgmetadata>
    <maintainer type="person">
        <email>martin-kokos@protonmail.com</email>
        <name>Martin Mokry</name>
    </maintainer>
	<upstream>
		<remote-id type="github">usebruno/bruno</remote-id>
		<maintainer type="person">
	        <email>anoop.md1421@gmail.com</email>
			<name>Anoop M D</name>
		</maintainer>
	</upstream>
</pkgmetadata>
```

MrRoy: (Even if they do use Gentoo, they could also be put in `<upstream>`)

martin-kokos: Updated.

stkw0:
I get the next error when executing: `[24785:0115/233152.475312:FATAL:setuid_sandbox_host.cc(157)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /opt/Bruno/chrome-sandbox is owned by root and has mode 4755.`.

I run a hardened profile so I am not sure if it is reproducible on non-hardened profiles. May be worth checking discord ebuild as that package also does all this fixes.

MrRoy:
@martin-kokos are you still interested in having this PR merged? If so can you please check stkw0's comment?

martin-kokos:
Yes, please.
I've bumped version and added `fperms +x` for `chrome-sandbox` and `chrome_crashpad_handler`

stkw0:
Thanks! Merged

PR #120, opened 2023-11-28
net-im/nheko: redesign ebuild

CyberTailor:
cc @tastytea


tastytea:
looks good to me, thank you! 💖

not sure the cmake `*_FOUND*` stuff is needed though, since it is supposed to be guaranteed to be there by the dependencies :thinking:

CyberTailor:
> not sure the cmake *_FOUND* stuff is needed though

Just the availability of cross-compiling always makes me paranoid.

CyberTailor:
Pushed

PR #122, opened 2023-12-09
www-plugins/firefoxpwa: add 2.9.1

filips123:


filips123:
I see that some improvements to the package have been added in https://github.com/gentoo/guru/commit/e87f462d5bce0ccd2378f54d5b83b41cfdff7dcd. Because this package is auto-generated from the source in my repository, those improvements are not available here.

I'll try to merge those changes into my repository and this PR.

filips123:
> Isn't "data-url" crate required anymore?

It now uses an official version from crates.io, so Git version is not needed anymore.

> Btw I think you forgot to add the licenses of crates to LICENSE

I'll fix that.

PR #124, opened 2024-01-15
https://bugs.gentoo.org/910572

Bratzmeister:
fixed rename from dev-util/cmake to dev-build/cmake


Bratzmeister:
also see dd1fefe6428552b0263aa3b23259c0a9ae40c792

CyberTailor:
We can't apply your commits without `Signed-off-by`

CyberTailor:
This needed urgent action, so I had to fix this tree-wide and this PR is no longer relevant

mgorny:
Thanks anyway.

PR #126, opened 2024-01-20
app-misc/cpufetch: add ~arm64 and ~ppc64 keywords

pkubaj:
Works fine on both.


stkw0:
Thanks! Merged to dev. Will be closed when it's merged to master.

PR #128, opened 2024-01-22
www-plugins/firefoxpwa: add 2.10.0

filips123:


filips123:
Due to a bug, I had to release another version, so this PR is replaced by #129.

PR #131, opened 2024-01-27
gui-apps/yambar: add ~arm64 keyword

pkubaj:
Works ok.


stkw0:
Thanks, merged. Just curious, did you try on arm64 hardware or in a QEMU virtualized system or something else?

Also, in case you are interested you can join GURU so you will be able to make those commits directly, see https://wiki.gentoo.org/wiki/Project:GURU/Information_for_Contributors

pkubaj:
arm64 hardware (PineBook Pro)

PR #133, opened 2024-02-06
dev-cpp/simdutf: new package, add 4.0.9

imwints:


stkw0:
Hi! Are you already registered to Gentoo bugzilla? 
Also, in case you didn't see it, you can file an access request to GURU to be able to commit directly. For more inormation see https://wiki.gentoo.org/wiki/Project:GURU/Information_for_Contributors

PR #139, opened 2024-02-22
net-im/abaddon: add missing spdlog dependency

ZerdoX-x:
https://github.com/uowuo/abaddon/blob/30b901b4df7344021ca79af4bccf95f6e799a37c/CMakeLists.txt#L117


micielski:
thanks!

PR #140, opened 2024-02-23
net-vpn/candy: new package with patch for ixwebsocket-11.4.4

YangMame:


micielski:
What is this patch accomplishing? Isn't this an upstream issue?

micielski:
pkgcheckscan throws some errors for net-vpn/candy:
 - MissingRemoteId: missing <remote-id type="github">lanthora/candy</remote-id> (inferred from URI 'https://github.com/lanthora/candy')
 - PkgBadlyFormedXml: package metadata.xml is not well formed xml: Opening and ending tag mismatch: maintainer line 10 and upstream, line 13, column 13 (metadata.xml, line 13)
 - BadFilename: version 3.9.1: bad filename: [ v3.9.1.tar.gz ]
 
 for your future reference: [pkgcheck page on gentoo wiki](https://wiki.gentoo.org/wiki/Pkgcheck)
Could you fix that?

Also I think it would make sense if you filed an access request to the repository as described in the wiki. For now, I believe I can merge your commit (after you will fix these warnings) since I see that you are on Gentoo's Bugzilla, so you should be able to receive notifications about bugs.

Thanks for contributing!!

micielski:
merged!

PR #143, opened 2024-02-25
Vaultwarden/Bitwarden cli, Vaultwarden server with web UI

rahilarious:


stkw0:
Seems two days ago they opened a PR to move from node:18 to node:20 https://github.com/bitwarden/clients/pull/8078, but given that the PR has no meaningful changes maybe currently it also works with node:20

rahilarious:
> Seems two days ago they opened a PR to move from node:18 to node:20 [bitwarden/clients#8078](https://github.com/bitwarden/clients/pull/8078), but given that the PR has no meaningful changes maybe currently it also works with node:20

Thank you for this. 

vaultwarden-web compiled fine with nodejs:20, so removed hard dependency on 18 in that.
Bitwarden-cli was half successful (building part) but generating package for Linux failed (electron part). Resulting in [error](https://github.com/vercel/pkg/issues/2014)

PR #144, opened 2024-02-26
dev-ml/*: add myself as maintainer to some packages

fogti:


stkw0:
I think you can remove the previous maintainer as he was no longer maintaining those packages. 

Pinging @MrRoy so he can exclude the now maintained packages from his PR

fogti:
cc @Alessandro-Barbieri (afaik previous maintainer)

MrRoy:
Thank you, I will remove them from the other PR when I can

stkw0:
@fogti I think you can safely remove it, those packages were meant to be removed anyway, and re-adding the previous maintainer will be easy if needed anyways.

stkw0:
Important point. For a package to be able to be keep and maintained their dependencies should also be maintained by someone. For example dev-ml/cstruct depends on dev-ml/crowbar, which is still masked.

fogti:
How can I filter the pkgcheck output to `dev-ml` or such (because otherwise it is pretty much unusable due to being too cluttered)

stkw0:
You can run `pkgcheck scan dev-ml` for example

dev-ml/cstruct/cstruct-6.2.0.ebuild
```diff
@@ -0,0 +1,85 @@
...
+	ppx? ( sexp )
+	test? ( sexp ppx )
+"
+
+src_prepare() {
+	default
+
+	if has_version '>=dev-lang/ocaml-5.1'; then
+		eapply "${FILESDIR}/6.2-ocaml-5.1.patch"
+	fi
```
fogti, line 63: ok, I think this might be unnecessary (as in: can always be applied, the dependency on ocaml-migrate-parsetree is redundant/unnecessary independent of ocaml version)

stkw0: I have literally no idea about this topic, nor ocaml in general, but seems like ocaml-migrate-parsetree has been superseded by ppxlib (https://ocaml.org/changelog/2023-10-23-omp-deprecation)?

fogti: it still requires that individual libraries need to be ported. In this case it appears that the dependency on ocaml-migrate-parsetree is just an outdated artifact from such a port.

stkw0:
I saw some of the commits don't have the Signed-off-by tag. https://wiki.gentoo.org/wiki/Project:GURU/Information_for_Contributors has more information about the different ways of adding the signoff tag more easily.

fogti:
This should be fixed now.

stkw0:
Thanks! Merged

stkw0:
Seems that there was a problem with the email. Did you follow the wiki and check that the email was the same on bugzilla? If you need help you can probably contact some dev on IRC

fogti:
I can rewrite the log to use the email from bugzilla

PR #145, opened 2024-02-29
net-vpn/candy: new 4.0 ebuild with version and cflags env patch

YangMame:
patch from upstream: https://github.com/lanthora/candy/commit/de8f5a3109ee3bdc472fbc7fcb1572aaad0b8cb9


stkw0:
Thanks. Merged

PR #150, opened 2024-03-10
Enable keyring feature of dev-libs/qtkeychain for net-im/nheko

jjdredd:
The requirement is mentioned in the nheko readme

### DIFF
--- a/net-im/nheko/nheko-0.11.3-r1.ebuild
+++ b/net-im/nheko/nheko-0.11.3-r1.ebuild
@@ -1,4 +1,4 @@
-# Copyright 2020-2023 Gentoo Authors
+# Copyright 2020-2024 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=8
@@ -26,7 +26,7 @@ COMMON_DEPEND="
 	dev-libs/libfmt:=
 	>=dev-libs/mtxclient-0.9.0:=
 	dev-libs/olm
-	>=dev-libs/qtkeychain-0.12.0:=[qt5]
+	>=dev-libs/qtkeychain-0.12.0:=[keyring,qt5]
 	dev-libs/spdlog:=
 	dev-qt/qtcore:5
 	dev-qt/qtdbus:5
--- a/net-im/nheko/nheko-0.11.3-r2.ebuild
+++ b/net-im/nheko/nheko-0.11.3-r2.ebuild
@@ -1,4 +1,4 @@
-# Copyright 2020-2023 Gentoo Authors
+# Copyright 2020-2024 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=8
@@ -26,7 +26,7 @@ COMMON_DEPEND="
 	dev-libs/libfmt:=
 	>=dev-libs/mtxclient-0.9.0:=
 	dev-libs/olm
-	>=dev-libs/qtkeychain-0.12.0:=[qt5]
+	>=dev-libs/qtkeychain-0.12.0:=[keyring,qt5]
 	dev-libs/spdlog:=
 	dev-qt/qtcore:5
 	dev-qt/qtdbus:5
--- a/net-im/nheko/nheko-9999.ebuild
+++ b/net-im/nheko/nheko-9999.ebuild
@@ -33,7 +33,7 @@ RDEPEND="
 	dev-libs/libfmt:=
 	dev-libs/olm
 	>=dev-libs/openssl-1.1.0:=
-	>=dev-libs/qtkeychain-0.14.1-r1:=[qt6]
+	>=dev-libs/qtkeychain-0.14.1-r1:=[keyring,qt6]
 	>=dev-libs/re2-0.2022.04.01:=
 	dev-libs/spdlog:=
 	>=dev-qt/kdsingleapplication-1.1.0:=[qt6]


tastytea:
i believe kwallet works without the flag? i would expect anyone who uses a secret agent that requires it to have it enabled globally.

MrRoy:
@jjdredd looks to me from tastytea's comment that this is not needed?

jjdredd:
> @jjdredd looks to me from tastytea's comment that this is not needed?

I didn't understand the tastytea's comment, but I do believe it's needed since it's a part of the nheko's dependency.
You can refer to the Nheko's readme:

`If you are using Gnome Keyring or KeepassXC as your secrets daemon, ensure that the keyring useflag is enabled on dev-libs/qtkeychain.`

https://github.com/Nheko-Reborn/nheko

MrRoy:
> `If you are using Gnome Keyring or KeepassXC as your secrets daemon, ensure that the keyring useflag is enabled on dev-libs/qtkeychain.`

So it means only if Gnome Keyring or KeepassXC is the secrets daemon, then the USE flag is needed, not if you use another secrets daemon such as kwallet

jjdredd:
> So it means only if Gnome Keyring or KeepassXC is the secrets daemon, then the USE flag is needed, not if you use another secrets daemon such as kwallet

It didn't mention kwallet. If that is so, do you suggest that I implement that condition in the ebuild? Would it break something if this use flag is enabled anyway? I suppose it could be done with a set of mutually exclusive use flags like `kwallet`, `keypass`, `gnome-keyring`, etc,  some of which trigger the `keyring` use.

MrRoy:
@tastytea would something like this make sense?
```
RDEPEND="
	...
	|| ( kde-frameworks/kwallet dev-libs/qtkeychain[keychain] )
	>=dev-libs/qtkeychain-0.14.1-r1:=[qt6]
	...
"
```

tastytea:
i don't understand how someone would get into the situation of using a secrets daemon that needs the flag and not have it enabled already?
also there is a message that you might need it when you install nheko:

```bash
optfeature "secrets storage support other than kwallet (for example gnome-keyring or keepassxc)" \
        "dev-libs/qtkeychain[keyring]"
```

jjdredd:
1. The secrets daemon doesn't need any flags enabled for qtkeychain, nheko needs it.
2. Don't assume everybody's born with the knowledge that the secrets daemon support needs to be enabled for all possible packages in case sometime in the future one needs to install an application that depends on those packages. Why does the user need to keep all the use flags in mind?
3. Even in the world where everyone knows at birth which gentoo use flags need to be enabled, there is no reason this task should not be automated.
4. The nheko message about keyring is not included in the normal ebuilds, just 9999

tastytea:
> 1. The secrets daemon doesn't need any flags enabled for qtkeychain, nheko needs it.

users of the relevant daemons need to have that flag enabled globally. not just for nheko. that's why it is enabled by default in gnome profiles.

> 2. Don't assume everybody's born with the knowledge that the secrets daemon support needs to be enabled for all possible packages in case sometime in the future one needs to install an application that depends on those packages and the specific secrets daemon. Why does the user need to keep all the use flags in mind? Why do gentoo ebuilds have the feature that allows to specify the required use flags for dependencies when everyone should remember them anyway?

it's only a required dependency for *some* users. people will complain if we add unnecessary dependencies.

> 5. The nheko message about keyring is not included in the normal ebuilds, just 9999

that's a bug, feel free to fix

> The idea of making the user manually keep track of dependencies and their features, when you have a package manager that's designed to do exactly that automatically, seems absurd to me.

the package manager is designed to give users a lot of flexibility. forced dependencies that are not necessary are working against that.

jjdredd:
> > 1. The secrets daemon doesn't need any flags enabled for qtkeychain, nheko needs it.
> 
> users of the relevant daemons need to have that flag enabled globally. not just for nheko. that's why it is enabled by default in gnome profiles.

Gnome secrets  used to work just fine for me without that use flag.
 
> > 2. Don't assume everybody's born with the knowledge that the secrets daemon support needs to be enabled for all possible packages in case sometime in the future one needs to install an application that depends on those packages and the specific secrets daemon. Why does the user need to keep all the use flags in mind? Why do gentoo ebuilds have the feature that allows to specify the required use flags for dependencies when everyone should remember them anyway?
> 
> it's only a required dependency for _some_ users. people will complain if we add unnecessary dependencies.

So why not enable it for _some_ users then? I think there's a different problem you're not telling me about.

> > 5. The nheko message about keyring is not included in the normal ebuilds, just 9999
> 
> that's a bug, feel free to fix

At this point it's easier to just stop using guru than fight some of the contributors to have your fixes accepted.

> > The idea of making the user manually keep track of dependencies and their features, when you have a package manager that's designed to do exactly that automatically, seems absurd to me.
> 
> the package manager is designed to give users a lot of flexibility. forced dependencies that are not necessary are working against that.

See above.

tastytea:
> See above. I am fine with @MrRoy proposition or something like that. Instead of arguing we could have worked out a better, sufficiently flexible solution.

what if there is a secrets daemon in a 3rd-party repo that doesn't need the flag? also i don't have the time to keep track of which daemons need the flag and which ones don't.

anyway, feel free to take over the package, but remove me from `metadata.xml`. no need to fight me, i'd actually be glad to have less work.

xz-dev:
As an aside, I think Gentoo's philosophy is that if you're not satisfied, Fork it.

MrRoy:
@jjdredd I think we can close this PR with the changes done in 1b40fd0917bdae0b06f5cc13a269f25d80b7e377?

jjdredd:
> @jjdredd I think we can close this PR with the changes done in [1b40fd0](https://github.com/gentoo/guru/commit/1b40fd0917bdae0b06f5cc13a269f25d80b7e377)?

You can close this pr because no progress is being made. The initial solution isn't good enough and there are no other proposals.
The changes in that mentioned commit just add a reminder to the user to apply a workaround themselves.
I also believe that these

```
optfeature "additional image formats support" kde-frameworks/kimageformats:5
	optfeature "audio & video file playback support" media-plugins/gst-plugins-meta[ffmpeg]
```
deserve separate use flags. I'm surprised that the author didn't add all the dependencies as a reminder to the user to install them manually.

tastytea:
> I'm surprised that the author didn't add all the dependencies as a reminder to the user to install them manually.

your snarkyness is tiring.

> The usage of a `USE` flag should not control runtime dependencies when the package does not link to it. Doing so will create extra configuration for the package and re-compilation for no underlying file change on disk. This should be avoided and instead can be conveyed to the user via post install messages if needed.
— <https://devmanual.gentoo.org/general-concepts/use-flags/#when-not-to-use-use-flags>

PR #157, opened 2024-03-26
*/*: update email on my packages

BratishkaErik:
Requested in https://bugs.gentoo.org/867043 .


BratishkaErik:
Why does `emails/bugzilla` check still fail...

PR #162, opened 2024-04-08
x11-wm/firefox-i3-workspaces: Fix description

foopsss:
Replaced two "s" letters that were written in upper-case with their lower-case counterparts.


MrRoy:
@foopsss before we can merge this you need to sign-off your commit using `git commit --signoff`
See: https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin

foopsss:
My apologies for the late reply. I've noticed that the typo was addressed in another commit already so this MR is no longer useful, although I'll set up the sign-off thing for future contributions.

PR #168, opened 2024-04-26
dev-util/bash-language-server: update SRC_URI

xz-dev:


MrRoy:
Please sign-off your commit
Also this should be revbumped unless the package is just broken and uninstallable without this?

CC: @tastytea

tastytea:
> Please sign-off your commit Also this should be revbumped unless the package is just broken and uninstallable without this?

it is uninstallable. apparently i uploaded the deps-tarball but forgot to include it in the ebuild :facepalm:

xz-dev:
> Please sign-off your commit Also this should be revbumped unless the package is just broken and uninstallable without this?
> 
> CC: @tastytea
> 
> Edit: no need to revbump

Thanks to remind

tastytea:
thank you! pushed to dev

PR #169, opened 2024-04-26
dev-libs/lsquic: treeclean again

antecrescent:
To my knowledge, lsquic has not been permitted back into the tree. I'm opening a PR because I'm not sure, if this is correct :)

The former maintainer added the package again after it was removed in 639a8db9b113179e86958249db5a9f49ed18e4f7 for depending on BoringSSL. Support for OpenSSL has not been added yet:
https://github.com/litespeedtech/lsquic/issues/113

Closes: https://bugs.gentoo.org/921749
Closes: https://bugs.gentoo.org/910246
Closes: https://bugs.gentoo.org/910096


MrRoy:
It is better to mask packages in `profiles/package.mask` (with an explanation) and then treeclean after some time has passed.

See also: https://github.com/gentoo/guru/commit/8e97eedc2eb0bc96302aca6027a7e323cdb992aa
It seems that it is linked statically so it does not have to be installed system-wide

antecrescent:
How could I forget, when I've personally masked a package before..
> It seems that it is linked statically so it does not have to be installed system-wide

Then, I don't have a problem with this package staying in ::guru. Sorry for the noise.

We may want to keep in mind that this package is quite outdated and seemingly unmaintained if someone decides to to a tree-wide cleanup in the future. It has had its last activity in July 2023 while the last upstream release happened this March.
I'll contact the maintainer to find out if they're still interested in this package and potentially drop them as a maintainer.

PR #173, opened 2024-04-30
gui-apps/lan-mouse: Install desktop,icon,service files

DMLhope:


gui-apps/lan-mouse/lan-mouse-0.7.3.ebuild
```diff
@@ -287,3 +294,10 @@ LICENSE+="
 "
 SLOT="0"
 KEYWORDS="~amd64"
+
+src_install() {
+	cargo_src_install
+	doicon -s scalable resources/de.feschber.LanMouse.svg
+	domenu de.feschber.LanMouse.desktop
```
stkw0, line 301: If installing desktop or icons you should call xdg_desktop_database_update and xdg_icon_cache_update in pkg_postinst and pkg_postrm. You can import xdg eclass for that.

DMLhope: Thank you very much for the reminder, I have added the corresponding function to the code.

stkw0: In this case I believe it's enough with inheriting xdg. Because no other eclass export pkg_post{inst,rm} you don't need to add those yourself

DMLhope: OK I deleted the extra pkg_postinst and pkg_postrm.

PR #186, opened 2024-05-08
app-misc/cmd-wrapped: new package, add 0.2.0

DMLhope:


stkw0:
In case you were not aware, you can register as a regular contributor, which allows you to commit directly without the extra step of making a PR and waiting for someone to merge it. See https://wiki.gentoo.org/wiki/Project:GURU/Information_for_Contributors

app-misc/cmd-wrapped/cmd-wrapped-0.2.0.ebuild
```diff
@@ -0,0 +1,85 @@
...
+SRC_URI="
+	${CARGO_CRATE_URIS}
+	https://github.com/YiNNx/cmd-wrapped/archive/refs/tags/${PV}.tar.gz -> ${P}.tar.gz
+"
+
+LICENSE="MIT"
+# Dependent crate licenses
+LICENSE+=" MIT MPL-2.0 Unicode-DFS-2016"
+SLOT="0"
+KEYWORDS="~amd64"
```
MrRoy, line 85: Rust packages need to have a `QA_FLAGS_IGNORED=` variable, otherwise the Gentoo Tinderbox (QA tool) will flag this ebuild as ignoring CFLAGS;

You can add something like
`QA_FLAGS_IGNORED="usr/bin/${PN}` (if this matches the binary path of course)

MrRoy (commented):
Hi @DMLhope,

Like stkw0 explained above, it's easy to get access to commit directly to the `dev` branch, which gets merged to `master` every once in a while.\
For new packages, we prefer people to obtain commit access directly rather than maintain the package through PRs, due to how frequent PRs may be required to properly maintain the package.

Otherwise the ebuild looks good except one minor thing

PR #188, opened 2024-05-12
fnm: use ring 17, add ppc64 support

darkbasic:


MrRoy:
LGTM, merged.
Have you reached out to upstream to update the ring crate? So eventually we can drop this patch?

PR #191, opened 2024-05-18
dev-util/hut: version bump to 0.5.0

cjbayliss:


cjbayliss:
As of yesterday, according to https://git.sr.ht/~emersion/hut, the maintainer and repo for hut has changed owner. It is now here: https://git.sr.ht/~xenrox/hut

Since there are no new releases, I don't think this PR should change.

dev-util/hut/hut-0.5.0.ebuild
```diff
@@ -21,7 +21,6 @@ SLOT="0"
 
 KEYWORDS="~amd64"
 
-
 DEPEND="${RDEPEND}"
```
antecrescent, line 24: Delete this line, since both `RDEPEND` and `DEPEND` are empty.

dev-util/hut/hut-0.5.0.ebuild
```diff
@@ -11,7 +11,7 @@ HOMEPAGE="https://sr.ht/~emersion/hut/"
 
 SRC_URI="
 	https://git.sr.ht/~emersion/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz
```
antecrescent, line 13: Please update this to reflect the upstream change.

dev-util/hut/hut-0.5.0.ebuild
antecrescent, line 1: - Fix the copyright, please.
 The package was introduced to ::guru last year and it should read `2023-2024`.
- Update `HOMEPAGE` to reflect the upstream change.
- If you decide to ship a pre-built manpage, remove the `doc/hut.1` target in `src_compile` and install the file in `src_install`.

antecrescent (changes requested):
> As of yesterday, according to https://git.sr.ht/~emersion/hut, the maintainer and repo for hut has changed owner. It is now here: https://git.sr.ht/~xenrox/hut
> 
> Since there are no new releases, I don't think this PR should change.

Since the upstream information in `metadata.xml` now needs changing, I suggest you tackle both that and this ebuild at once, so that they are up to date and in sync :)

Changes for `metadata.xml`:
- Remove `<remote-id type="github">apraga/hut-vendor</remote-id>`
 It's not related to upstream.
- Update `<remote-id type="sourcehut">~emersion/hut</remote-id>` to reflect the new upstream.

Thank you for bumping this package!

dev-util/hut/hut-0.5.0.ebuild
```diff
@@ -21,7 +21,6 @@ SLOT="0"
 
 KEYWORDS="~amd64"
 
-
 DEPEND="${RDEPEND}"
 BDEPEND="app-text/scdoc"
```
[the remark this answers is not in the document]
eli-schwartz: This is incorrect. The policy states that:
- `USE=man` must not be used, since the package must always install a manpage
- prebuilding it and shipping it as an additional source is a *good alternative* to requiring "inconvenient" additional dependencies

scdoc isn't an inconvenient additional dependency. I don't think it's necessary to prebuild it, as the package currently doesn't violate the policy document.

cjbayliss:
Thank you for the review!! :heart: 

I tried dropping the dependency on `scdoc`, and ran into an issue where the man page was greater than 20KB. Since I didn't know how to deal with that, I asked on `#gentoo-dev-help`. They suggested adding the file to `SRC_URI` instead of the repo.

Then they suggested that depending on `scdoc` wasn't a bad idea since it isn't hard to install. I agree with them, so I have not dropped `scdoc`.

If you insist on it, I'll drop `scdoc` and host the man page in the same repo I'm using the the go modules. :slightly_smiling_face:

MrRoy:
LGTM, merged in https://github.com/gentoo/guru/commit/027c1e52cdbbaf017c44be48104345c14e1082db

dev-util/hut/hut-0.5.0.ebuild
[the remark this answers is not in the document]
antecrescent: @eli-schwartz It seems I misunderstood part of the policy. Your second bullet point clarified it for me, thanks!

PR #192, opened 2024-05-21
media-plugins/dtmediawiki: add that images always will be encoded

vitaly-zdanevich:


vitaly-zdanevich:
Ooops, I this is only a mirror.

PR #196, opened 2024-06-03
net-vpn/candy: add openrc init script

irort:
add openrc init script for candy, please check it @YangMame


YangMame (approved):

YangMame:
It's better to add a README elog

PR #197, opened 2024-06-07
dev-build/just-1.28.0: fix install

xz-dev:


MrRoy:
@joecool1029 FYI

MrRoy:
@xz-dev seems to be fixed by https://github.com/gentoo/guru/commit/56a211875e7decddc116b10194464fb62ad8c4e8

joecool1029:
sorry yeah, I came up with a proper fix. The one time I didn't test it before bumping (seriously I've bumped this package like 20 times with no drama)

joecool1029:
if you want to improve upon what I have that's welcome though, but it's not a solution to just not generate the manpages and completitions, we should ship them.

PR #201, opened 2024-06-25
net-vpn/mullvadvpn-app: Add gui to /usr/bin

justin025:
The previous ebuild installed the cli components to /usr/bin but did not install the gui, only a desktop file. The changes below install the gui application alongside the cli components instead of having to manually create a symbolic link.

The sed portion is necessary as the launch script does not work unless inside the application directory, in this case /opt/Mullvad VPN. By specifying the 'SCRIPT_DIR' the application launches fine.


stkw0:
I'm not sure if for so few lines it's needed, but just in case, could you add the certificate of origin as per policy? See https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin

For your information, you can register yourself to GURU as a contributor, which will grant you permissions to commit directly to the dev branch. See https://wiki.gentoo.org/wiki/Project:GURU/Information_for_Contributors

justin025:
I've added a certificate of origin to the commit, would you prefer I close this pr and commit to the dev branch instead?

stkw0:
As you wish. It's banned to commit to master directly, I can merge it to dev.
To merge this PR, please first fixup all commits so there is only a single commit instead of three.

justin025:
Apologies, I've opened a new pr here https://github.com/gentoo/guru/pull/202

PR #203, opened 2024-06-25
net-im/gtkcord4: remove package

gentoo-setan:
It's been renamed to dissent and it's at version 0.25 now.
https://github.com/diamondburned/dissent/
This is badly outdated and I think should be removed


stkw0:
Thank you very much for the report. 
It can not be removed directly. First it has to be masked in order to make users aware of the removal and being able to give them time to adapt, change to an alternative or so. 

Given you are the current maintainer for the package, do you want to handle it @Schievel1?

Schievel1:
Yes, I agree. I couldn't update it anymore because it didn't build for me with portage. 

Since this is ::guru and not ::gentoo, do we still need to stick to the 30 day masking procedure?

antecrescent:
> Since this is ::guru and not ::gentoo, do we still need to stick to the 30 day masking procedure?

Yes, we also stick to in ::guru.

Schievel1:
I just pushed 4e513c2. The removal is at the end of August because of my vacation in between.

PR #206, opened 2024-06-29
Version bump sys-kernel/ugrd to 1.14.0 and add option feature for nilfs2

oblivikun:
I dont know if option feature is the right name for optfeature

### DIFF
--- a/profiles/package.mask
+++ b/profiles/package.mask
@@ -18,6 +18,13 @@
 
 # New entries go on top.
 
+# Pascal Jaeger <pascal.jaeger@leimstift.de> (2024-06-29)
+# Package is renamed to 'dissent'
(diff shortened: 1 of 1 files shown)


stkw0:
Please use the standard git commit message (sys-kernel/ugrd: add 1.14.0 ....), pkgdev can help with that.

oblivikun:
Sorry for the newbie question, but how do I change the git message after I committed something

stkw0:
You can do `git commit --amend`  and then force-push (which in general is something that should not be done, but for a PR in a branch that only you will commit is acceptable). 
For general guidelines and tools to make the workflow easier see https://wiki.gentoo.org/wiki/Project:GURU/Information_for_Contributors. It also has information about how to request commit access in case you would want to contribute more often.

PR #214, opened 2024-07-13
media-sound/spotify-adblock: Add USE "savedconfig"

justin025:
Adds savedconfig functionality to customize allow / deny domain list. Upstream is a bit permissive by default.


antecrescent:
According to the repo's README you can configure the allow- and denylist `/etc/spotify-adblock/config.toml`. Does that not work?

justin025:
I guess you're right, i suppose an elog directing the user to /etc/spotify-adblock/config.toml might make more sense.

PR #215, opened 2024-07-15
acct-user/ollama: simple tweak for the daemon to work

CreeperMain:
@vitaly-zdanevich @antecrescent Hello again, this time I'm submitting a simple tweak because of the ollama daemon, it will not run the `ollama serve` command if the ollama user's shell is set to /sbin/nologin. I change it so it uses /bin/false because the official docs recommend that and voila the daemon works as intended
https://github.com/ollama/ollama/blob/main/docs/linux.md#adding-ollama-as-a-startup-service-recommended

### DIFF
--- a/acct-user/ollama/ollama-1.ebuild
+++ b/acct-user/ollama/ollama-1.ebuild
@@ -8,6 +8,7 @@ inherit acct-user
 DESCRIPTION="A user for ollama"
 ACCT_USER_ID=-1
 ACCT_USER_HOME=/var/lib/ollama
+ACCT_USER_SHELL=/bin/false
 ACCT_USER_HOME_PERMS=0700
 ACCT_USER_GROUPS=( ollama )
 


antecrescent:
Please file a bug over at https://bugs.gentoo.org/enter_bug.cgi?product=GURU&component=Package+issues with instructions on how to reproduce this issue and link to it here.
I just spun up an empty container and the following worked just fine (running as `root`):
```sh
# emerge -1a app-misc/ollama
[...]
# rc-service ollama start
 * Caching service dependencies ...                                                                                                                                                                            [ ok ]
 * Starting Ollama Service ...
# ollama list
NAME    ID      SIZE    MODIFIED
```

CreeperMain:
@antecrescent Thank you I will

antecrescent:
@vitaly-zdanevich Could you look into packaging `app-misc/ollama-v0.2.x`, please? `9999`-only packages are discouraged because they are extremely unstable and harder to debug than fixed versions :3

vitaly-zdanevich:
@antecrescent without live ebuild we cannot use [go-module_live_vendor](https://devmanual.gentoo.org/eclass-reference/go-module.eclass/index.html)? Do I need to host dependencies in my repo?

antecrescent:
> @antecrescent without live ebuild we cannot use [go-module_live_vendor](https://devmanual.gentoo.org/eclass-reference/go-module.eclass/index.html)? Do I need to host dependencies in my repo?

Yes, you need to create a dependency tarball and host it somewhere.

CreeperMain:
I'll close this pr, I will file another bug report that involves this package and submit the fix as for this pr the issue was on my end it was related to the ollama group

PR #221, opened 2024-07-19
dev-php/fpdf: new package, add 1.86

vitaly-zdanevich:
Copied from https://github.com/bobwya/bobwya/tree/master/dev-php/fpdf, fixed a little.

I need this package as a dependency for versioned [app-benchmarks/phoronix-test-suite](https://github.com/gentoo/guru/tree/master/app-benchmarks/phoronix-test-suite) - will create PR after.

The license looks like a shorter MIT - what do you think?


jayofdoom:
This seems OK to me. Package looks to have an MIT license.

dev-php/fpdf/metadata.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
```
redawl, line 3: I know these aren't required, but it could be good to add upstream, and yourself as a maintainer.

dev-php/fpdf/fpdf-1.86.ebuild
```diff
@@ -0,0 +1,33 @@
...
+
+DESCRIPTION="FPDF is a PHP class which allows to generate PDF files with pure PHP"
+HOMEPAGE="http://www.fpdf.org/"
+SRC_URI="http://www.fpdf.org/en/dl.php?v=${MY_PV}&f=tgz -> ${MY_P}.tgz"
+
+S="${WORKDIR}/${MY_P}"
+
+LICENSE="freedist"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
```
redawl, line 18: Has this been tested on x86? I believe keywords should only be added for arches that the package has been tested on, even for unstable keyword.

vitaly-zdanevich: Deleted `~x86`

dev-php/fpdf/metadata.xml
vitaly-zdanevich: Thanks, done.

vitaly-zdanevich:
> This seems OK to me. Package looks to have an MIT license.

Thanks - changed `freedist` to `MIT`.

eli-schwartz:
It is technically not an MIT license -- the MIT license imposes a condition upon you, the condition of attribution. The MIT-0 license does NOT impose this condition, and is closer to the fpdf license -- though the fpdf license has modified wording and doesn't disclaim liability.

One could argue that this software really *is* the MIT-0 license with meaningless fat trimmed off. I'm not a lawyer and cannot answer whether that argument holds water.

I don't think you can argue this is the MIT not-zero license, though. :)

PR #224, opened 2024-08-05
licenses: add MIT-fpdf

vitaly-zdanevich:
For https://github.com/gentoo/guru/pull/221


stkw0:
It should be fpdf. If it's not an MIT license why should it be called "MIT-fpdf"?

antecrescent:
This commit should be part of https://github.com/gentoo/guru/pull/221. There's no need for an additional PR.

vitaly-zdanevich:
ok

eli-schwartz:
> It should be fpdf. If it's not an MIT license why should it be called "MIT-fpdf"?

To be fair, it does appear to be an MIT derivative. So the name doesn't come out of nowhere.

PR #231, opened 2024-08-15
www-apps/redlib: fix passing address in the OpenRC init file

tokoyami:
ADDRESS is not supported as an environment variable currently.  It has to be passed as an argument.

PORT has been changed to a normal variable and passed as argument to make the init file similar to the SystemD service file where both ADDRESS and PORT are passed as arguments.

### DIFF
--- a/www-apps/redlib/files/redlib.initd
+++ b/www-apps/redlib/files/redlib.initd
@@ -3,8 +3,8 @@
 # Distributed under the terms of the GNU General Public License v2
 
 # Environment variables for Redlib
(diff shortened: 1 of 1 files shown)


stkw0:
It looks good. It's missing the Signed-off-by, though. See https://wiki.gentoo.org/wiki/Project:GURU/Information_for_Contributors

tokoyami:
Sorry about that.  Added now.

I force-pushed out of habit.  I hope it's not going to be an issue.

stkw0:
Thanks. Don't worry. Force pushing for PR is fine

PR #236, opened 2024-09-06
Update packettracer to 8.2.2

Xephobia:
Recently, Cisco changed their _netacad_ platform and broke the old 8.2 packet tracer. This adds the updated packet tracer and fixes the issue.


stkw0:
btw it will be more clean to fixup  or squash the add "missing desktop entry" commit and have only a single commit.

PR #238, opened 2024-09-08
media-sound/nulloy: fix theme adding

vitaly-zdanevich:


media-sound/nulloy/nulloy-0.9.9.ebuild
```diff
@@ -8,9 +8,10 @@ inherit xdg
 DESCRIPTION="Music player with a waveform progress bar (sound visualization)"
 HOMEPAGE="https://nulloy.com"
 
-NAME="nulloy-theme-night-v1.0"
+THEME_VERSION=1.1.1
+NAME="nulloy-theme-night-${THEME_VERSION}"
 SRC_URI="https://github.com/nulloy/nulloy/archive/refs/tags/${PV}.tar.gz -> ${P}.tar.gz
-	skins? ( https://gitlab.com/vitaly-zdanevich/nulloy-theme-night/-/archive/v1.0/${NAME}.tar.bz2 )"
+	skins? ( https://gitlab.com/vitaly-zdanevich/nulloy-theme-night/-/archive/${THEME_VERSION}/${NAME}.tar.bz2 )"
```
stkw0, line 14: I think will make more sense to make a PR to include the theme

vitaly-zdanevich: Hm... lets merge this pull request for now please...

jayofdoom: I am not a fan either of mixing sources here; we should let users choose if they want the upstream nulloy release separate from if they want your additional theme.

jayofdoom (changes requested):
I would strongly prefer we not require all users of nulloy to pull an unofficial theme from an different github organization to enable any theming at all. IMO it belongs in a separate package, or, at the *absolute minimum* a non-default-enabled use flag.

vitaly-zdanevich:
Ok, I will move my theme to a separate ebuild.

PR #240, opened 2024-09-12
sys-power/auto-cpufreq: add 2.4.0

xz-dev:


stkw0:
This was already on the tree

PR #241, opened 2024-09-13
media-sound/nulloy: drop external theme

vitaly-zdanevich:
I will move it to a separate ebuild (decided by the community).


stkw0:
Please remove the skins use flag. Afterward, you can use optfeature to inform users of optional packages that adds more skins. While on that may be a good idea to remove old versions.

vitaly-zdanevich:
Now `skins` flag decides whether to install integrated upstream skins.

stkw0:
Perfect. I missed the `$(use skins || echo --no-skins)` line and thought it was no longer used-

PR #243, opened 2024-09-15
Drop ia64 keyword

johu:
See https://www.gentoo.org/news/2024/08/14/Gentoo-drops-IA-64-support.html


stkw0:
In this case I think you can put all drops in one single commit like `*/*: drop ~ia64 keywrod` instead of one commit per package.

stkw0:
Seems it has been already merged in https://github.com/gentoo/guru/commit/5dc8172b794480b26dcb9b51231559a4d5e2c0ce

PR #246, opened 2024-09-26
www-plugins/firefoxpwa: add 2.12.5

filips123:


stkw0:
Thanks. From time to time, please clean older versions.

filips123:
How old versions should be removed? I guess all other than the latest few?

stkw0:
That depens on the criteria of the maintainer. Personally, I usually leave the newer version and one older version

PR #250, opened 2024-10-14
dev-cpp/wt: add pango useflag, enabled by default.

PPN-SD:
Allow to reduce dependencies. Pango may not be necessary for some use cases of wt.


stkw0:
You missed changing  `-DENABLE_PANGO=ON` to `-DENABLE_PANGO=$(usex pango)`

PPN-SD:
Fixed. A description was included in the metadatas as pango is not a global useflag.

PR #253, opened 2024-10-28
sys-fs/dwarfs: fix missing dev-cpp/range-v3 dependency

Xephobia:
Closes: https://bugs.gentoo.org/940265


stkw0:
Create a -r1 for this fix as it touches he run time dependencies.

stkw0:
Also the linked issue doesn't seems to have any relation with the fix¿?

Xephobia:
> Also the linked issue doesn't seems to have any relation with the fix¿?

see my [comment](https://bugs.gentoo.org/940265#c3)
TLDR : upstream clones dependencies if they are missing, which can't happen when building with `FEATURES=network-sandbox` (a default) because it prevents network access. In this case range-v3 was missing and it cannot even resolve the github.com domain because it doesn't have any network, so no DNS.

PR #259, opened 2024-11-09
sys-auth/hyprpolkitagent: new package

johu:


MrRoy:
Thanks, merged.
I added a version check for GCC/Clang since this doesn't build on GCC13.
See 44d3efff43db84a327ff24e9aa9e95b9077d71a3

PR #262, opened 2024-11-18
Fix symbola-13.00.ebuild SRC_URI

bananas-repos:
This fixes the broken SRC_URI to a working webarchive URL.

But this does not solve the up-to-dateness of this ebuild, which I can not answer.

### DIFF
--- a/media-fonts/symbola/symbola-13.00.ebuild
+++ b/media-fonts/symbola/symbola-13.00.ebuild
@@ -9,7 +9,7 @@ MY_PN="${PN^}"
 
 DESCRIPTION="Unicode font for Latin, IPA Extensions, Greek, Cyrillic and many Symbol Blocks"
 HOMEPAGE="https://dn-works.com/ufas/"
-SRC_URI="https://dn-works.com/wp-content/uploads/2020/UFAS-Fonts/Symbola.zip -> ${P}.zip"
+SRC_URI="https://web.archive.org/web/20220327073316/https://dn-works.com/wp-content/uploads/2020/UFAS-Fonts/Symbola.zip -> ${P}.zip"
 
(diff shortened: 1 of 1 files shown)


PPN-SD:
New PDFs have been upload recently : https://dn-works.com/wp-content/uploads/UFAS/
I think we need to extract fonts from PDF with fontforge from now on.

bananas-repos:
Manually and add it as a source or while package install?

PPN-SD:
Ah, we can't with new files from website ...
Last version with otf embedded is 14 from : https://web.archive.org/web/20240107144224/https://dn-works.com/wp-content/uploads/2021/UFAS121921/Symbola.pdf

Debian source, version 13, could be a good and stable alternative but it's only ttf, not otf : http://deb.debian.org/debian/pool/main/t/ttf-ancient-fonts/ttf-ancient-fonts_2.60.orig.tar.xz

bananas-repos:
archlinux does have the same problem. https://aur.archlinux.org/packages/ttf-symbola

It looks like an licence issue: https://aur.archlinux.org/packages/ttf-symbola#comment-993205

> Please do not flag this as out of date for v15 or v16 unless the licence situation changes and a download is available. The upstream developer has changed the licensing from "free for personal use but pay for academic and commercial use" to "pay for any use". Additionally there is no longer a publicly available download, so there is nothing to update this package too

See: https://dn-works.com/wp-content/uploads/UFAS/License.pdf

PPN-SD:
I made a new PR #263 to replace this package for ttf-ancient-fonts with stable source, more fonts and no issue with license.

bananas-repos:
Thx. I will close this, since it is not needed anymore.

PR #272, opened 2025-01-02
Update clapper-0.6.1.ebuild

rvlobato:
It was having complaints about missing IUSE app flag


antecrescent:
Hi,

thanks for your contribution. This bug is already recorded at https://bugs.gentoo.org/947321. Let's keep the discussion in one place, please :)

Best
Lucio

rvlobato:
> Hi,
> 
> thanks for your contribution. This bug is already recorded at https://bugs.gentoo.org/947321. Let's keep the discussion in one place, please :) Your patch also needs to add a corresponding description for the `app` USE flag in `metadata.xml` and update the copyright notice to `2023-2025`.
> 
> I believe, this would be a good use case for the global USE flag `gui` instead of `app`.
> 
> Best Lucio

Thanks for pointing out the record :)

PR #274, opened 2025-01-07
Bump vesktop-bin to v1.5.4

VincentAhluwalia:
Added manifest and new ebuild file. Nothing changed between versions, just a simple bump.


stkw0:
For the future, please use pkgdev to ensure that the commits follows the standard commit message and ensure that it's signed off. Thsi time given it was a trivial change I merged it directly myself.

PR #275, opened 2025-01-12
Hyprpolkitagent 0.1.2

johu:


stkw0:
The sys-auth package is missing the metadata.xml file.

johu:
> The sys-auth package is missing the metadata.xml file.

Because its already in git.

stkw0:
Missed that the Manifest was being edited not added. Sorry

PR #279, opened 2025-01-15
www-plugins/firefoxpwa: add 2.13.3

filips123:


stkw0:
Can at least the two older versions be removed? Usually, old versions are pruned from time to time.
If I recall correctly, this PR was automated in the CI/CD, right? In that case it might make sense to remove the oldest version in the same workflow.

filips123:
To remove old versions, removing their ebuilds and regenerating the `Manifest` with `pkgdev manifest` should be enough, right? I will probably add something to CI for future versions to remove all but the latest few, or something similar.

For this version, can you please do this, as I don't have time currently?

PR #282, opened 2025-01-23
gui-apps/sfwbar: Add USE flags

VincentAhluwalia:
Added USE flags to make all of the modules optional since they were enabled by default. E.g not everyone needs bluetooth support, or simultaneous iwd AND networkmanager support. 

No errors with pkgcheck and it is compiling without errors. Added respective USE flag descriptions as well.

Removed xdg from inherit because it wasn't used anymore.

Pull request because I'm not sure if maintainer @LBCrion agrees. Tried to reach per E-Mail but no luck.

### DIFF
--- a/gui-apps/sfwbar/metadata.xml
+++ b/gui-apps/sfwbar/metadata.xml
@@ -6,7 +6,16 @@
 		<name>Lev Babiev</name>
 	</maintainer>
 	<use>
-		<flag name="mpd">Add support for Music Player Daemon</flag>
+		<flag name="mpd">Enable support for Music Player Daemon</flag>
+		<flag name="menu">Enable support for application menu generator. Default enabled by upstream</flag>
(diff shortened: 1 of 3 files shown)


LBCrion:
Apologies, I haven't received your email (it likely got eaten by an overzealous spam filter)
I'm happy to have these flags added. Even if these modules don't introduce additional dependencies,  it makes sense to make them optional.

VincentAhluwalia:
Ok, great! I added back the xdg-utils icon cache stuff because portage complained. Also when I disable some features (e.g notification center) the notification center is still being displayed. Not a big problem because it is technically disabled but it could confuse some users. Do I need to change something in the ebuild or is it an upstream issue?

VincentAhluwalia:
Hey, since you agree I'm gonna close this pr and commit. Faster that way.

PR #283, opened 2025-01-25
gui-apps/hyprsysteminfo: new package, add 0.1.3

johu:


stkw0:
I have seen you have already packages on the tree. You are already registered on GURU, right? This PR was made to wait the review of someone specifically?

johu:
> I have seen you have already packages on the tree. You are already registered on GURU, right? This PR was made to wait the review of someone specifically?

I am a former Gentoo developer, but as I know I have no write access to any Gentoo repos since 2022 anymore.

stkw0:
In that case, requesting access as a normal user should allow you to have push access as any other user, right? The general policy is to not merge PR for new packages. This seems to be a kind of special case, but still.

johu:
I don't see this as a requirement in https://wiki.gentoo.org/wiki/Project:GURU#The_regulations

stkw0:
About not merging PR for new package: that's correct, it has not been written down. Pull Requests are not, the official way to contribute so there is no rules about that, though it has been discussed multiple times. Informally, contributors can merge some PR and so, but no one is expected to do it.

johu:
I have access now. https://bugs.gentoo.org/949864 Will setup my git repo now and then push to the dev branch.

PR #285, opened 2025-01-31
Update libsfdo-0.1.3.ebuild

nopicnick:
fix meson config


demize:
This easiest way to get your changes into ::guru is to request access and commit them yourself; check out [the wiki](https://wiki.gentoo.org/wiki/Project:GURU) for instructions on how to request access and how to properly format your commit message for GURU. We don't use PRs here much, so it's probably faster to request access and push it yourself :)

As for the change itself, I believe this would require a rev bump (libsfdo-0.1.3 -> libsfdo-0.1.3-r1) as the existing build didn't fail completely. A rev bump will force the package to be rebuilt by anyone with it installed.

PR #287, opened 2025-02-13
dev-zig/zls: switch to slotted zig

demize:
Opening a PR for this because I'm not convinced it's actually a good idea, but a discussion with @BratishkaErik is probably worth having. Marking it as a draft since it's more for feedback anyway, and I'm sure there's changes to be made if this is something that makes sense.

Trying to get myself started with zig (and specifically trying to get a couple variations of a project off the ground) I'm finding myself switching between zig master and zig 0.13. But I also noticed that ZLS pretty clearly warns in its [install instructions](https://zigtools.org/zls/) that you should match your version of ZLS with the version of zig you're using.

So I threw this together, as what feels like a somewhat idiomatic approach (given zig is in ::gentoo and ZLS is in ::guru). It's based heavily on the approach in the zig ebuild itself, and while I'm not really a fan of a separate `eselect zls` here, since I'd rather see it integrated into `eselect zig`, this does seem to work.

But like I said, I'm not convinced it's a good idea, so I'd appreciate any feedback (or criticism!) from Eric or anyone else. In any case, it's a major change, and not one I'd make on my own (on my package, and *especially* not on anyone else's).

Feel free to ping me on IRC to discuss; I'm demize95 on there :)

### DIFF
--- a/app-eselect/eselect-zls/eselect-zls-1.ebuild
+++ b/app-eselect/eselect-zls/eselect-zls-1.ebuild
@@ -1,0 +1,20 @@
+# Copyright 2023-2024 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+DESCRIPTION="Manages Zig Language Server versions"
+HOMEPAGE="https://wiki.gentoo.org/wiki/No_homepage"
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~arm64"
+
+RDEPEND="app-admin/eselect"
+
+S="${WORKDIR}"
+
+src_install() {
+	insinto /usr/share/eselect/modules/
+	newins "${FILESDIR}"/zls.eselect-${PVR} zls.eselect
+}
--- a/app-eselect/eselect-zls/files/zls.eselect-1
+++ b/app-eselect/eselect-zls/files/zls.eselect-1
@@ -1,0 +1,132 @@
+# -*-eselect-*-  vim: ft=eselect
+# Copyright 2022-2024 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+DESCRIPTION="Manage Zig Language Server versions"
+MAINTAINER="demize@unstable.systems"
+
+show_selected_target() {
+	readlink "${EROOT}/usr/bin/zls"
+}
+
+find_targets() {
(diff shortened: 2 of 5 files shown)


BratishkaErik:
I'm not sure I understand utility of this. Do you have some problems with using ZLS 9999 with Zig 0.13 or similar mix? Maybe some problems with `zls.build.json`?

Even if server itself reports some warnings like

> [coc.nvim] ZLS 0.14.0-dev.384+360bd96 does not support Zig 0.13.0. A ZLS 0.13 release should be used instead.

it works very well (at least for me) in practice (else I wouldn't be able to make updates from 0.13 to latest while maintaining autocomplete support for both), apart from some rare cases like not suggesting old builtins. 

I just tested and it works reasonably well even with 0.10:

изображение

If you have some issues with the current setup I'll gladly help with this PR. Otherwise, I believe the added complexity is not justified IMHO.

> But like I said, I'm not convinced it's a good idea

BratishkaErik:
In other words, did you encounter an issue yourself, or is this change solely to align the ebuild with upstream recommendations?

demize:
It's primarily to try and match the upstream recommendation.

I don't have any very significant codebases to test on at the moment, but I do see at least one issue in my (so far very minimal) project. If I'm using zig master and ZLS 0.13, the same build.zig.zon file gives me an error from ZLS reading the file (`zls: file cannot be a tuple`).

I've heard from other people that ZLS 0.13 works fine with zig master, so I think it's the sort of issue that's likely to depend on multiple conditions, and may crop up for some people and not others; which is exactly the sort of reason I'd prefer to match the upstream recommendation.

But I do generally agree with you that it's added complexity, and probably not justified here unless something breaks more significantly. Makes more sense to stick to ZLS master across zig versions. I'll close this PR, feel free to reuse anything I did here if it proves to be necessary.

BratishkaErik:
Thanks for your explanation.

> If I'm using zig master and ZLS 0.13, the same build.zig.zon file gives me an error from ZLS reading the file (zls: file cannot be a tuple).

I see, IIRC it was AstGen error and fixed on latest ZLS, and prevented format-on-save? I disabled this functionality long time ago (when contributing to ncdu) and call `zig fmt` manually, so sadly I don't remember how this part works with different Zig versions.

> unless something breaks more significantly. Makes more sense to stick to ZLS master across zig versions.

Good luck :+1:

PR #291, opened 2025-02-20
*/*: disable py3.10

CyberTailor:
Python 3.10 target is scheduled for removal in Gentoo:
https://public-inbox.gentoo.org/gentoo-dev/e6ca9f1f4ababb79ab2a3d005c39b483c3ecef6e.camel@gentoo.org/

Sphinx 8.2.0 already dropped support for CPython 3.10 and PyPy 3.10,
leading to multiple CI failures in ::guru.

We can safely remove this target, because GURU is "experimental" and
does not have stable keywords.

Let's not remove "pypy3" right now because
a) There are not many packages using it.
b) So we can see which packages supported PyPy 3.10, and test them with
   PyPy 3.11.


gui-apps/organicmaps/organicmaps-9999.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI=8
 
-PYTHON_COMPAT=( python3_{7..12} )
+PYTHON_COMPAT=( python3_{7,11,12} )
```
ceamac, line 6: Why keep 7?

x11-misc/i3-restore/i3-restore-4.1.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI=8
 
-PYTHON_COMPAT=( python3_{7..12} )
+PYTHON_COMPAT=( python3_{7,11,12} )
```
ceamac, line 6: Ditto

dev-python/mwclient/mwclient-0.10.1.ebuild
```diff
@@ -4,7 +4,7 @@
 EAPI=8
 
 DISTUTILS_USE_PEP517=setuptools
-PYTHON_COMPAT=( python3_{5..12} )
+PYTHON_COMPAT=( python3_{5,6,7,11,12} )
```
ceamac, line 7:
```suggestion
PYTHON_COMPAT=( python3_{11..12} )
```

app-backup/pika-backup/pika-backup-0.7.4-r1.ebuild
```diff
@@ -322,7 +322,7 @@ CRATES="
 	zvariant_utils@1.0.1
 	zxcvbn@2.2.2
 "
-PYTHON_COMPAT=( python3_{9..13} )
+PYTHON_COMPAT=( python3_{11,12,13} )
```
ceamac, line 325:
```suggestion
PYTHON_COMPAT=( python3_{11..13} )
```

app-containers/waydroid-images/waydroid-images-9999.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI=8
 
-PYTHON_COMPAT=( python3_{8..12} )
+PYTHON_COMPAT=( python3_{11,12} )
```
ceamac, line 6:
```suggestion
PYTHON_COMPAT=( python3_{11..12} )
```

app-emulation/cloud-utils/cloud-utils-0.33.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI=8
 
-PYTHON_COMPAT=( python3_{8..13} )
+PYTHON_COMPAT=( python3_{11,12,13} )
```
ceamac, line 6:
```suggestion
PYTHON_COMPAT=( python3_{11..13} )
```

demize:
> b) So we can see which packages supported PyPy 3.10, and test them with PyPy 3.11.

Do we have a PyPy 3.11 target we can use yet? `pypy3` is being dropped from packages in ::gentoo simultaneously with `python3_10`, but I’m not sure if we have the `pypy3_11` target yet.

For packages that get broken by their dependencies, I’d like to be able to test with pypy3.11, though I’m not suggesting it has to be done as part of this.

gui-apps/organicmaps/organicmaps-9999.ebuild
CyberTailor: I used `gpy-impl $f -python3_{8..10}` and missed some extreme cases

CyberTailor:
> > b) So we can see which packages supported PyPy 3.10, and test them with PyPy 3.11.
> 
> Do we have a PyPy 3.11 target we can use yet? `pypy3` is being dropped from packages in ::gentoo simultaneously with `python3_10`, but I’m not sure if we have the `pypy3_11` target yet.
> 
> For packages that get broken by their dependencies, I’d like to be able to test with pypy3.11, though I’m not suggesting it has to be done as part of this.

Yes, but it's necessary to make sure test suite passes with PyPy 3.11 first.

CyberTailor:
Pushed v2 with suggestions from ceamac

CyberTailor:
Oops, accidentally pushed the old version to dev

CyberTailor:
> Oops, accidentally pushed the old version to dev

Pushed changes as a separate commit.

PR #294, opened 2025-02-27
+ zen-bin-1.8.2b.ebuild

dualfade:
zen-bin latest; please review.

Thank you.


stkw0:
As you prefer. In theory, the correct way to contribute is to register as per wiki instructions and commit directly to the dev branch.

dualfade:
Right  I see it. Ok let me drop this; re pull and do this the correct way.

Thanks !

PR #296, opened 2025-02-28
app-doc/geeknote: add 3.0.6

vitaly-zdanevich:
This update includes only this useful commit https://github.com/vitaly-zdanevich/geeknote/commit/578bfcfda0b35f65ab7ab1575030303bfa39d9fb


stkw0:
Why are tests restricted even when they are enabled?

vitaly-zdanevich:
> there is no DEPENDS?

I have dependencies in RDEPEND, because geeknote needs them to Run, no?

stkw0:
I was not sure if it was needed in case of enabling tests , but seems it is not.

PR #301, opened 2025-03-11
games-roguelike/tsl new package

vitaly-zdanevich:
[Migrated](https://github.com/leycec/raiagent/tree/master/games-roguelike/tsl) from the repo that will be [archived soon](https://github.com/leycec/raiagent/issues/117).

The official website was unavailable - so I found the code in another place and put it into my repo.

I have two warnings, what to do with them?

>  * QA Notice: Files built without respecting CFLAGS have been detected
>  *  Please include the following list of files in your report:
>  * /usr/bin/tsl-tty

>  * QA Notice: Files built without respecting LDFLAGS have been detected
>  *  Please include the following list of files in your report:
>  * /usr/bin/tsl-tty


stkw0:
> The official website was unavailable - so I found the code in another place and put it into my repo.

It's not reasonable to merge this when upstream is dead.

vitaly-zdanevich:
But the game is working. If you do not want - you can close, and I will add to another overlay.

stkw0:
There is a couple of considerations that are important. It's not about if I want or not, it's about if it's right or not, for which we can have a conversation about it if we disagree, but I think this case is pretty clear:

0) A package should be feasible to maintain
1) This is NOT free software, the License only allows patches or ports, but seems they are not happy with a fork
2) Upstream is dead, so there is no way to report issues, fix them or get any help

In my opinion, this by itself is enough reason for not merging the package, but there are other potential issues:
- The source code don't seem to come from a trustworthy source. Specially given that upstream is dead
- Even if they were happy with a fork, it may require maintaining the source code, and if I recall correctly from other threads, you didn't knew C. An important part of the source code seems to be written in C

PR #303, opened 2025-03-15
.github/workflows: replace changed-files action with git-diff

pastalian:
tj-actions/changed-files repo was compromised and has been removed.

### DIFF
--- a/.github/workflows/emails.yml
+++ b/.github/workflows/emails.yml
@@ -8,6 +8,10 @@ on:
   pull_request:
     branches: [dev]
 
+defaults:
+  run:
+    shell: bash
+
 jobs:
   bugzilla:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/emails.yml
```diff
@@ -40,17 +40,13 @@ jobs:
...
       - name: Get changed files
         if: ${{ fromJSON(env.is_commit_valid) }}
-        uses: tj-actions/changed-files@v44
-        with:
-          base_sha: ${{ env.last_successful_commit }}
-          write_output_files: true
-          files: |
-            **/metadata.xml
+        run: |
+          git diff --diff-filter d --name-only ${last_successful_commit} -- '*/metadata.xml' | tr '\n' ' ' | tee all_changed_files.txt
```
thesamesam, line 44: Do you need to set pipefail or something here, or does Actions handle that?

pastalian: Looks like we need to specify `shell: bash` for that.
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#defaultsrunshell

thesamesam (approved):

pastalian:
Thanks for the review!

PR #310, opened 2025-04-02
net-wireless/sdrpp: allow rtl-sdr-blog to satisfy dependency

TimovVeen:
net-wireless/rtl-sdr-blog is a drop in replacement fork of net-wireless/rtl-sdr recently added to the gentoo tree. This change allows either one to be used with sdr++
@redawl


redawl:
Looks good to me. Are you able to push to dev? I don't have merge permission for github.

TimovVeen:
Not yet, but I have requested permission. I'll push it when it gets accepted, unless someone comes along and merges it before that.

PR #312, opened 2025-04-05
www-client/zen/bin: correct Manifest byte size

Sneethe:
www-client/zen-bin-1.11b byte size wasn't updated from zen-bin-1.10.3b value


TimovVeen:
The hash is different too, so this doesn't work. I'll update the manifest and push it to dev directly

stkw0:
I this kind of cases, it's better if the maintainer (in this case @MrRoy or Michal Vu, which I don't know the gh handle)  takes a look to ensure that upstream tarball was not modified and if it was to investigate why.

TimovVeen:
Understood, my apologies. In this case I read from the upstream discord that the original download had issues which is why it was reuploaded, so it was nothing malicious.

stkw0:
My point was to not just update the Manifest blindly, but if you already knew the reason it's all good :) Next time though, it would be still better to add that in the commit description.

stkw0:
The fix has already been merged. This PR can be closed.

Sneethe:
>The hash is different too, so this doesn't work. I'll update the manifest and push it to dev directly

I really need to stop doing late night PRs

> My point was to not just update the Manifest blindly, but if you already knew the reason it's all good :) Next time though, it would be still better to add that in the commit description.

Good point. Thanks for the headsup.

PR #313, opened 2025-04-07
dev-zig/zls: improve ebuild a bit

fkobi:
Moved from https://github.com/gentoo/guru/commit/1261a8a8e32a2f2a822592d1089a577088373701#r154919684.

### DIFF
--- a/dev-zig/zls/zls-0.14.0.ebuild
+++ b/dev-zig/zls/zls-0.14.0.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=8
 
-DESCRIPTION="The officially unofficial Ziglang language server"
+DESCRIPTION="Non-official language server for Zig"
 HOMEPAGE="https://zigtools.org/zls/ https://github.com/zigtools/zls"
 
 declare -g -r -A ZBS_DEPENDENCIES=(
@@ -12,14 +12,18 @@ declare -g -r -A ZBS_DEPENDENCIES=(
 	[N-V-__8AABhrAQAQLLLGadghhPsdxTgBk9N9aLVOjXW3ay0V.tar.gz]='https://github.com/ziglibs/diffz/archive/ef45c00d655e5e40faf35afbbde81a1fa5ed7ffb.tar.gz'
 )
 
-# Sync with "minimum_build_zig_version" from upstream's "build.zig".
 if [[ ${PV} == 9999 ]]; then
 	ZIG_SLOT="9999"
 
 	EGIT_REPO_URI="https://github.com/zigtools/zls"
 	inherit git-r3
+	src_unpack() {
+		git-r3_src_unpack
+		zig_live_fetch -Denable-tracy=false
+	}
 else
-	ZIG_SLOT="0.14"
+	# Should be the "minimum_build_zig_version" from upstream's "build.zig".
+	ZIG_SLOT="$(ver_cut 1-2)" # works only for releases, but that's okay
 
 	SRC_URI="
 		https://github.com/zigtools/zls/archive/refs/tags/${PV}.tar.gz -> zls-${PV}.tar.gz
@@ -42,15 +46,6 @@ RDEPEND="
 "
 
 DOCS=( README.md )
-
-src_unpack() {
-	if [[ ${PV} == 9999 ]]; then
-		git-r3_src_unpack
-		zig_live_fetch -Denable-tracy=false
-	else
-		zig_src_unpack
-	fi
-}
 
 src_configure() {
 	local my_zbs_args=(


stkw0:
Mostly stylistic changes, mostly removing the usage of variables with hardcoded versions. I think the version of the original author is fine as it is.
I don't agree with removing subslots as it may be useful and if not, doesn't hurt.
I agree with adding the GitHub to HOMEPAGE and the change to description also makes sense.  

Overall, it's controversial if this could be an improvement or not. If you stick to the last point, I think it can make sense, but for the rest I don't think should be touched.

stkw0:
Ah wait, I diffed them in the opposite direction. Then all my comments apply but negated (:

stkw0:
Reading the original thread, the maintainer seems that disagreed on using variables instead of hardcoded values and against adding the subslot. While in those points I agree with you, I think if the maintainer stated that he works better the way it's currently implemented and didn't replied to this PR for over a month, it's okay to leave it as it and close this PR. Or ping the maintainer if you think it important enough to discuss it further.

stkw0:
About 1, I didn't understand what do you mind with header bar.  In any case, I think it's useful to have the Github and HOMEPAGE and it's a standard approach. Removing it just makes people life (very slightly) harder without any benefit. 

About 2, I'm not native English speaker, but it sounds very weird when reading without the "The"

stkw0:
Depending on the case, I think it's more correct. As a totally random TM example, sys-apps/portage have as a description `The package management and distribution system for Gentoo`. 

About the HOMEPAGE, yes, it's not needed, but sometimes it's nice to have. There is absolutely no need to remove it. As an example, for pypi packages it's also one link away and as pypi is pretty standard it's very easy to know where to find it. Despite that, a lot of python packages have the two HOMEPAGEs, pypi and github

stkw0:
Yes, it's redundant, though redundancy is not always bad. The cost of maintaining it does not increase. It doesn't hurt in any way. If there was not there, I would not care. But removing it when the original writer of the ebuild and maintainer though it was useful (and I also think that) is unnecessary and not an improvement. 

As you pointed before and now with the reference to the devmanual, if there was only a single HOMEPAGE and from that webpage it was hard to get the source, adding a second page would be and improvement. If the source code can be easily found and there are two HOMEPAGEs, removing one is not an improvement. It is just a triviality up to the maintainer to decide.

BratishkaErik:
```patch
diff --git a/dev-zig/zls/zls-0.14.0.ebuild b/dev-zig/zls/zls-0.14.0.ebuild
index c37929d1e..b5eeaf24f 100644
--- a/dev-zig/zls/zls-0.14.0.ebuild
+++ b/dev-zig/zls/zls-0.14.0.ebuild
@@ -3,8 +3,8 @@
 
 EAPI=8
 
-DESCRIPTION="The officially unofficial Ziglang language server"
-HOMEPAGE="https://zigtools.org/zls/ https://github.com/zigtools/zls"
+DESCRIPTION="Officially unofficial Ziglang language server"
+HOMEPAGE="https://zigtools.org/zls/"
 
 declare -g -r -A ZBS_DEPENDENCIES=(
 	[known_folders-0.0.0-Fy-PJtLDAADGDOwYwMkVydMSTp_aN-nfjCZw6qPQ2ECL.tar.gz]='https://github.com/ziglibs/known-folders/archive/aa24df42183ad415d10bc0a33e6238c437fc0f59.tar.gz'
@@ -12,18 +12,16 @@ declare -g -r -A ZBS_DEPENDENCIES=(
 	[N-V-__8AABhrAQAQLLLGadghhPsdxTgBk9N9aLVOjXW3ay0V.tar.gz]='https://github.com/ziglibs/diffz/archive/ef45c00d655e5e40faf35afbbde81a1fa5ed7ffb.tar.gz'
 )
 
-# Sync with "minimum_build_zig_version" from upstream's "build.zig".
 if [[ ${PV} == 9999 ]]; then
-	ZIG_SLOT="9999"
+	ZIG_SLOT="${PV}"
 
 	EGIT_REPO_URI="https://github.com/zigtools/zls"
 	inherit git-r3
 else
-	ZIG_SLOT="0.14"
+	# Sync with "minimum_build_zig_version" from upstream's "build.zig".
+	ZIG_SLOT="$(ver_cut 1-2)" # works only for releases, but that's okay
 
-	SRC_URI="
-		https://github.com/zigtools/zls/archive/refs/tags/${PV}.tar.gz -> zls-${PV}.tar.gz
-	"
+	SRC_URI="https://github.com/zigtools/zls/archive/refs/tags/${PV}.tar.gz -> ${P}.tar.gz"
 	KEYWORDS="~amd64"
 fi
 
@@ -31,13 +29,14 @@ inherit zig
 SRC_URI+="${ZBS_DEPENDENCIES_SRC_URI}"
 
 LICENSE="MIT"
-SLOT="0"
+SLOT="0/${ZIG_SLOT}"
 
 # Sync with "minimum_runtime_zig_version" from upstream's "build.zig".
+runtime_ver="0.14.0"
 RDEPEND="
 	|| (
-		>=dev-lang/zig-0.14.0
-		>=dev-lang/zig-bin-0.14.0
+		>=dev-lang/zig-${runtime_ver}
+		>=dev-lang/zig-bin-${runtime_ver}
 	)
 "
 
```

#

 ```diff
-DESCRIPTION="The officially unofficial Ziglang language server"
+DESCRIPTION="Officially unofficial Ziglang language server"
```

I don't remember from where I got this description, I can't find it anywhere now... Website and `zls --help` both call themselves as "A non-official language server for Zig", so we can use this.

> "A " suggests that it's one of many. Why would you need that information?

Because it's an unofficial server, and there exist another servers which support Zig LSP, for example https://github.com/llogick/zigscient. This was one of the reasons why they changed style to use ZLS or "... language server for Zig" everywhere instead of "Zig Language Server", which can imply being from core developers.

```diff
-HOMEPAGE="https://zigtools.org/zls/ https://github.com/zigtools/zls"
+HOMEPAGE="https://zigtools.org/zls/"
```

https://projects.gentoo.org/qa/policy-guide/ebuild-format.html#pg0104

> Rationale: HOMEPAGE permits multiple entries by design, and developers are generally encouraged to add more helpful entries (e.g. project pages on PyPI, GitHub…). ...

And I like these links in `eix`... but if you really insist...

```diff
-	SRC_URI="
-		https://github.com/zigtools/zls/archive/refs/tags/${PV}.tar.gz -> zls-${PV}.tar.gz
-	"
+	SRC_URI="https://github.com/zigtools/zls/archive/refs/tags/${PV}.tar.gz -> ${P}.tar.gz"
```

I find it less readable, let's try to repeat what Flowdalic did in 06ba1f0f7:

```diff
 	SRC_URI="
-		https://github.com/zigtools/zls/archive/refs/tags/${PV}.tar.gz -> zls-${PV}.tar.gz
+		https://github.com/zigtools/zls/archive/refs/tags/${PV}.tar.gz
+			-> ${P}.tar.gz
 	"
```

Subslot changes: unneccessary complication

Dependency versioning: maybe let's try to group them if you want to extract? Something like this:

```bash
# Sync with upstream's build.zig
if [[ ${PV} == 9999 ]]; then
    minimum_ver_build="9999"
    minimum_ver_run="0.14.0"
else
    minimum_ver_build="0.14.0"
    minimum_ver_run="0.14.0"
fi

ZIG_SLOT="${minimum_ver_build}"

# ...

RDEPEND="
    || (
        >=dev-lang/zig-${minimum_ver_run}
        >=dev-lang/zig-bin-${minimum_ver_run}
    )
"
```

dev-zig/zls/zls-0.14.0-r1.ebuild
```diff
@@ -0,0 +1,61 @@
...
+RDEPEND="
+	|| (
+		>=dev-lang/zig-${runtime_ver}
+		>=dev-lang/zig-bin-${runtime_ver}
+	)
+"
+
+DOCS=( README.md )
+
+src_unpack() {
```
BratishkaErik, line 45: This can be improved to be like Flowdalic's logic from 06ba1f0f7 (again :) ):

```bash
if [[ ${PV} == "9999" ]]; then
	src_unpack() {
		git-r3_src_unpack
		zig_live_fetch -Denable-tracy=false
	}
fi
```

BratishkaErik:
It says about variable substitution, not content:
> If both overlap, the common part must be repeated verbatim.

And also: https://projects.gentoo.org/qa/policy-guide/other-metadata.html#pg0702
> The homepage specified in ebuilds is normally used to locate information about the upstream project, e.g. downloads, source code repository, bug tracker, documentation.

BratishkaErik (approved):

BratishkaErik:
> > I like these links in eix
> 
> Visually it's about the same as `emerge -s`. Is this output what you're referring to?
Yes

> I think neither `eix`, `emerge --search` nor `esearch` (visually identical to emerge btw) are the right tools for the job. Even the eix [readme](https://github.com/vaeth/eix/?tab=readme-ov-file#other-search-utils) recommends "other _search-utils_".

I think they just mention other projects for completeness, because in paragraph above they say "Provides more information than most other search utilities". When I'm not at PC for example, sometimes I also use gpo.zugaina.org, and it shows homepages without `remote-id` too.

dev-zig/zls/zls-0.14.0.ebuild
```diff
@@ -12,14 +12,18 @@ declare -g -r -A ZBS_DEPENDENCIES=(
 	[N-V-__8AABhrAQAQLLLGadghhPsdxTgBk9N9aLVOjXW3ay0V.tar.gz]='https://github.com/ziglibs/diffz/archive/ef45c00d655e5e40faf35afbbde81a1fa5ed7ffb.tar.gz'
 )
 
-# Sync with "minimum_build_zig_version" from upstream's "build.zig".
 if [[ ${PV} == 9999 ]]; then
 	ZIG_SLOT="9999"
 
 	EGIT_REPO_URI="https://github.com/zigtools/zls"
 	inherit git-r3
+	src_unpack() {
```
thesamesam, line 20: Please do _not_ do this. Conditionally defining phases is unusual and we don't want to see more of it. It also means all phases aren't in the same place.

thesamesam: Really, what you need to ask is:
* Is the style I'm proposing to change to unusual? How many instances are there in tree?
* Is the existing style in the ebuild wrong or that unusual? If not, you should keep it as-is and leave it to the current maintainer. Making changes like this implies that there's some canonical style you're changing to, but that's not what you're doing here.

thesamesam:
> I assume that's so that people do not set HOMEPAGEs as project's readmes.

No, the QA policy you link to talks about using the variable ${HOMEPAGE} _as a reference_.

dev-zig/zls/zls-0.14.0.ebuild
fkobi: This was not my idea and I do not like it.
I only implemented this because the maintainer suggested so: https://github.com/gentoo/guru/pull/313#discussion_r2082559065.

thesamesam: You are the person ultimately responsible for the changes you commit. But you could have at least asked whether you needed to fix it or if the maintainer was going to, as bare minimum?

Further, my remarks here apply to your other work too. This seems to be a pattern: not understanding critique as applying in general and to be transferred across PRs and packages.

fkobi: > you could have

Yes I could. I assumed too many things.

> You are the person ultimately responsible for the changes you commit.

This is an wise and important sentence and a lesson learnt for me.

PR #315, opened 2025-04-25
gui-wm/river: add 0.3.9

robertgzr:
fixes crashing when clients use the security context protocol, notably flatpak
see https://codeberg.org/river/river/releases/tag/v0.3.9


demize (approved):
Thanks!

Simple version bump, looks fine to me.

PR #316, opened 2025-04-28
add package openlitespeed

KotoishiHeart:
What to do if you don't have php and eselect-php support?


www-servers/openlitespeed/openlitespeed-1.8.3.1.ebuild
```diff
@@ -0,0 +1,85 @@
+# Copyright 1999-2024 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit cmake
+
+BORINGSSL_COMMIT="9fc1c33e9c21439ce5f87855a6591a9324e569fd"
+
+DESCRIPTION="LiteSpeed QUIC (LSQUIC) Library"
```
stkw0, line 10: Does not seems to be the correct description

www-servers/openlitespeed/openlitespeed-1.8.3.1.ebuild
```diff
@@ -0,0 +1,85 @@
...
+HOMEPAGE="https://github.com/litespeedtech/lsquic/"
```
stkw0, line 11: The homepage is also wrong

www-servers/openlitespeed/openlitespeed-1.8.3.1.ebuild
```diff
@@ -0,0 +1,85 @@
...
+SRC_URI="
+	https://github.com/litespeedtech/openlitespeed/archive/refs/tags/v${PV}.tar.gz -> ${P}.tar.gz
+	https://github.com/google/boringssl/archive/${BORINGSSL_COMMIT}.tar.gz -> boringssl-9fc1c3.tar.gz
+"
+
+LICENSE="MIT"
```
stkw0, line 17: The license is not correct

www-servers/openlitespeed/openlitespeed-1.8.3.1.ebuild
```diff
@@ -0,0 +1,85 @@
...
+	sed "s:%LSWS_CTRL%:$SERVERROOT/bin/lswsctrl:" "${S}/dist/admin/misc/lshttpd.service.in" > "${S}/dist/admin/misc/lshttpd.service"
+
+	local mycmakeargs=(
+	)
+	cmake_src_configure
+}
+
+src_install() {
+	default
+	if use systemd; then
```
stkw0, line 80: Both, systemd service and OpenRC one should be installed unconditionally

www-servers/openlitespeed/openlitespeed-1.8.3.1.ebuild
```diff
@@ -0,0 +1,85 @@
...
+	OPENLSWS_ADMINPORT=7080
+	DEFAULT_TMP_DIR=/tmp/lshttpd
+	if use ruby ; then
+		RUBY_PATH=/usr/bin/ruby
+	else
+		RUBY_PATH=
+	fi
+	OPENLSWS_EXAMPLEPORT=8088
+
+	sed -e "s/%ADMIN_PORT%/$OPENLSWS_ADMINPORT/" "${S}/dist/admin/conf/admin_config.conf.in" > "${S}/dist/admin/conf/admin_config.conf"
```
stkw0, line 67: Check potential errors  like you do with "mv"

KotoishiHeart:
> Can you explain a bit more what do you mean? What are you trying to achieve and how is openlitespeed related?

Openlitespeed requires a special php that supports litespeed.
The currently available PHP is not compatible with Litespeed.
The default PHP provided by openlitespeed is version 5.6 with no support.

demize:
> Openlitespeed requires a special php that supports litespeed.

It looks like PHP has support for litespeed, but it needs to be enabled at compile-time. The best solution would be to submit a separate pull request to ::gentoo, adding a `litespeed` USE flag so that this package can depend on `dev-lang/php[litespeed]`.

Unfortunately I don't think the alternative solutions to provide PHP are good fits for GURU, since they would mean either creating an entirely separate PHP ebuild in ::guru solely to support litespeed, bundling a custom build of PHP with openlitespeed, or expecting the user to provide their own PHP installation.

Theoretically you could add openlitespeed without providing any PHP support, but I'd probably try to get a USE flag added in ::gentoo first, since otherwise it leaves potential users in an awkward situation here.

www-servers/openlitespeed/openlitespeed-1.8.3.1.ebuild
```diff
@@ -0,0 +1,85 @@
...
+DESCRIPTION="LiteSpeed QUIC (LSQUIC) Library"
+HOMEPAGE="https://github.com/litespeedtech/lsquic/"
+SRC_URI="
+	https://github.com/litespeedtech/openlitespeed/archive/refs/tags/v${PV}.tar.gz -> ${P}.tar.gz
+	https://github.com/google/boringssl/archive/${BORINGSSL_COMMIT}.tar.gz -> boringssl-9fc1c3.tar.gz
+"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~arm64"
```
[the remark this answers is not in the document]
KotoishiHeart: Since openlitespeed supports ARM, I don't think there's any problem. I confirmed that I can build it.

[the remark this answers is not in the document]
stkw0: This is a common confusion. Even if upstream supports it,if it has not been personally tested it should not be added.  In this case it's okay since you tested it.

KotoishiHeart:
As a workaround until litespeed support is added, would it be better to install the official lsphp8.1 package for ubuntu?

demize:
> As a workaround until litespeed support is added, would it be better to install the official lsphp8.1 package for ubuntu?

Ultimately this is just a webserver, so I don’t think we should block adding it because there’s no easy way to build support for it with PHP in ::gentoo.

I don’t think we can add or include our own PHP server in ::guru, either as part of this package or as its own package, since it would conflict with the duplicates policy. And I don’t want to hold this up myself, so if we can add it without any PHP support (unless the user installs a custom build), I think that’s fine for now. Did you (or are you planning to) open a PR in ::gentoo to add the USE flag?

Either way, @stkw0 has provided a lot more feedback already, so if he doesn’t have any objections to adding as-is then don’t worry about me :)

stkw0:
I think it's worthy to send a PR to ::gentoo adding the litespeed support, as it may be useful for other cases and would simplify things. Mixing a source ebuild and a pre-built binary from Ubuntu is a bit of a mess.

btw the current PR still has a number of issues (will add some more comments). There is no sign-off tag and there are two Metadata files for acct-user and acct-group ebuilds that should not be there. 
Is there a need to add both, openlitespeed version 1.8.1 and version 1.8.3.1?

www-servers/openlitespeed/files/openlitespeed-disable-build-deps-libs.patch
```diff
@@ -0,0 +1,153 @@
...
+-endif()
+-if (CMAKE_HOST_SYSTEM_NAME MATCHES "Darwin")
+-    set (NOT_MACOS OFF)
+-else()
+-    set (NOT_MACOS ON)
+-endif()
+-OPTION(MOD_PAGESPEED "Whether pagespeed should be built" ${LINUX_X64})
+-OPTION(MOD_SECURITY "Whether mod_security should be built" ${NOT_MACOS})
+-OPTION(MOD_LUA "Whether Lua should be built" ${LINUX_X64})
++OPTION(MOD_PAGESPEED "Whether pagespeed should be built" OFF)
```
stkw0, line 26: this should be passed to mycmakeargs instead of being patched here

www-servers/openlitespeed/files/openlitespeed-add-install-files.patch
```diff
@@ -0,0 +1,84 @@
+I plan to eventually change the patch to install openlitespeed in the appropriate location instead of installing it in /usr/local/lsws/.
```
stkw0, line 1: yeah. This should be fixed before merging this

KotoishiHeart:
> Is there a need to add both, openlitespeed version 1.8.1 and version 1.8.3.1?

There's no need. It works only with Version 1.8.3.1.

> Did you (or are you planning to) open a PR in ::gentoo to add the USE flag?

I don't have the code knowledge to open a PR to add the litespeed flag to "eselect-php" or "php", and I don't have the time to do the work for that.
It might be possible if you do it for a long time, but other developers might be faster to change it.

stkw0:
> There's no need. It works only with Version 1.8.3.1.

Then please, remove the other one.



> I don't have the code knowledge to open a PR to add the litespeed flag to "eselect-php" or "php", and I don't have the time to do the work for that.
> It might be possible if you do it for a long time, but other developers might be faster to change it.

I just opened a PR to add litespeed USE to php: https://github.com/gentoo/gentoo/pull/42003

www-servers/openlitespeed/openlitespeed-1.8.3.1.ebuild
```diff
@@ -0,0 +1,82 @@
...
+	app-arch/brotli
+	dev-libs/yajl
+	virtual/libcrypt
+	ruby? ( dev-lang/ruby )
+	dev-libs/lsquic:=[static-libs=]
+	dev-libs/ls-qpack:=[static-libs=]
+	dev-libs/ls-hpack:=[static-libs=]
+"
+
+RDEPEND="
```
thesamesam, line 36: Several libraries are missing from RDEPEND here.

www-servers/openlitespeed/openlitespeed-1.8.3.1.ebuild
```diff
@@ -0,0 +1,82 @@
...
+	https://github.com/litespeedtech/openlitespeed/archive/refs/tags/v${PV}.tar.gz -> ${P}.tar.gz
+	https://github.com/google/boringssl/archive/${BORINGSSL_COMMIT}.tar.gz -> boringssl-9fc1c3.tar.gz
+"
+
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~arm64"
+
+IUSE="static-libs ruby systemd"
+RESTRICT=""
```
thesamesam, line 22: ?

www-servers/openlitespeed/openlitespeed-1.8.3.1.ebuild
```diff
@@ -0,0 +1,82 @@
...
+
+DEPEND="
```
thesamesam, line 24: Unsorted contents. Also, conditionals inside should go last.

www-servers/openlitespeed/openlitespeed-1.8.3.1.ebuild
```diff
@@ -0,0 +1,82 @@
...
+	fi
+	OPENLSWS_EXAMPLEPORT=8088
+
+	sed -e "s/%ADMIN_PORT%/$OPENLSWS_ADMINPORT/" "${S}/dist/admin/conf/admin_config.conf.in" > "${S}/dist/admin/conf/admin_config.conf" || die
+	sed -e "s/%USER%/$OPENLSWS_USER/" -e "s/%GROUP%/$OPENLSWS_GROUP/" -e "s#%DEFAULT_TMP_DIR%#$DEFAULT_TMP_DIR#"  -e "s/%ADMIN_EMAIL%/$OPENLSWS_EMAIL/" -e "s/%HTTP_PORT%/$OPENLSWS_EXAMPLEPORT/" -e "s/%RUBY_BIN%/$RUBY_PATH/"  "${S}/dist/conf/httpd_config.conf.in" > "${S}/dist/conf/httpd_config.conf" || die
+	sed "s:%LSWS_CTRL%:$SERVERROOT/bin/lswsctrl:" "${S}/dist/admin/misc/lsws.rc.in" > "${S}/dist/admin/misc/lsws.rc" || die
+	sed "s:%LSWS_CTRL%:$SERVERROOT/bin/lswsctrl:" "${S}/dist/admin/misc/lsws.rc.gentoo.in" > "${S}/dist/admin/misc/lsws.rc.gentoo" || die
+	sed "s:%LSWS_CTRL%:$SERVERROOT/bin/lswsctrl:" "${S}/dist/admin/misc/lshttpd.service.in" > "${S}/dist/admin/misc/lshttpd.service" || die
+
+	local mycmakeargs=(
```
thesamesam, line 73: Why?

www-servers/openlitespeed/openlitespeed-1.8.3.1.ebuild
```diff
@@ -0,0 +1,82 @@
...
+	)
+	cmake_src_configure
+}
+
+src_install() {
+	default
```
thesamesam, line 79: This should surely be `cmake_src_install`.

www-servers/openlitespeed/openlitespeed-1.8.3.1.ebuild
```diff
@@ -0,0 +1,82 @@
...
+	systemd_dounit "${D}"/admin/misc/lshttpd.service || die
```
thesamesam, line 80: ED, not D, but no need for die here (it dies for us).

www-servers/openlitespeed/openlitespeed-1.8.3.1.ebuild
```diff
@@ -0,0 +1,82 @@
...
+IUSE="static-libs ruby systemd"
+RESTRICT=""
+
+DEPEND="
+	dev-lang/go
+	dev-libs/modsecurity
+	app-arch/brotli
+	dev-libs/yajl
+	virtual/libcrypt
+	ruby? ( dev-lang/ruby )
```
thesamesam, line 30: This is unlikely to be right by itself and should use an eclass.

thesamesam (changes requested):

PR #318, opened 2025-05-07
sys-fs/dwarfs: add 0.12.3-r1, drop 0.12.3

mhx:
- added `cpptrace` dependency for `stacktrace` use flag
- removed unnecessary libdir tweaking
- added `-g` compile flag for builds with `stacktrace` use flag


sys-fs/dwarfs/dwarfs-0.12.3-r1.ebuild
```diff
@@ -60,12 +60,8 @@ RESTRICT="!test? ( test )"
 CHECKREQS_DISK_BUILD="500M"
 CMAKE_WARN_UNUSED_CLI=0
 
-src_prepare(){
-	cmake_src_prepare
-	sed "s/DESTINATION lib/DESTINATION $(get_libdir)/" -i cmake/libdwarfs.cmake || die
-}
-
 src_configure(){
+	use stacktrace && append-flags -g
```
eli-schwartz, line 64: I assume the idea behind this is that configuring an exception formatter doesn't do anything if you don't build with DWARF debuginfo?

sys-fs/dwarfs/dwarfs-0.12.3-r1.ebuild
```diff
@@ -60,12 +60,8 @@ RESTRICT="!test? ( test )"
 CHECKREQS_DISK_BUILD="500M"
 CMAKE_WARN_UNUSED_CLI=0
 
-src_prepare(){
-	cmake_src_prepare
-	sed "s/DESTINATION lib/DESTINATION $(get_libdir)/" -i cmake/libdwarfs.cmake || die
```
eli-schwartz, line 65: Thanks for following up on this. I guess the commit message can explain that it used to be necessary, but isn't *anymore*, and reference https://github.com/mhx/dwarfs/commit/ca69d2b141356b4cbb8d16d2607bd002faaf121a ?

sys-fs/dwarfs/dwarfs-0.12.3-r1.ebuild
eli-schwartz: And I see that cpptrace does support GNU debuglink, so this should work well with gentoo's `FEATURES="splitdebug"` (by default it will just strip all that info).

mhx: > I assume the idea behind this is that configuring an exception formatter doesn't do anything if you don't build with DWARF debuginfo?

There will still be stack traces even without debug symbols, but they'll require post-processing to be meaningful as they'll only contain addresses.

> And I see that cpptrace does support GNU debuglink, so this should work well with gentoo's `FEATURES="splitdebug"` (by default it will just strip all that info).

Yes, I've actually tested that this works fine with `splitdebug`.

sys-fs/dwarfs/dwarfs-0.12.3-r1.ebuild
mhx: Done, wasn't sure how much detail was needed :)

stkw0:
Thank you very much for the PR. 
Could you add the Signed-off-by tag (see https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin)?

mhx:
> Thank you very much for the PR. Could you add the Signed-off-by tag (see https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin)?

Sure, done.

PR #322, opened 2025-05-19
dev-python/pympress: new package, add 1.8.6

APN-Pucky:


dev-python/pympress/pympress-1.8.6.ebuild
```diff
@@ -0,0 +1,40 @@
+EAPI=8
+
+PYTHON_COMPAT=( python3_{11..13} )
```
pastalian, line 3: https://github.com/Cimbali/pympress/issues/330#issuecomment-2830496819
It seems that Python 3.13 with the current `dev-python/pygobject` version (3.50.0) doesn't work without a patch.

APN-Pucky: Ah, thanks for reviewing! I will remove `python3_13` for now then and add a comment linking to the issue.

APN-Pucky: Adding the patch turned out to be fairly easy and it works with `python3_13` now.

PR #323, opened 2025-05-21
games-emulation/duckstation: new package, add 0.1.8947

kwhat:

PR #324, opened 2025-05-21
dev-libs/libbacktrace: new package, add 1.0_p20250410

kwhat:
This is being added because it is a required dependency of [games-emulation/duckstation](https://github.com/gentoo/guru/pull/323).


vimproved:
Just a note: from experience of packaging this in a personal overlay, this package needs a filter-lto.

libbacktace defines a `BACKTRACE_SUPPORTED` macro in backtrace-supported.h that gets set to zero if it determines the library will not work correctly (but it will still compile for some reason!). In the process of settings this macro, it greps in an object file which fails with ThinLTO due to object files just being LTO bitcode.

PR #328, opened 2025-06-01
app-shells/direnv: Add EPREFIX to emake command

Jimmy2027:
This allows to install direnv with a custom prefix


Jimmy2027:
ping @demize :)

PR #330, opened 2025-06-05
games-util/minigalaxy: undo edo

PPN-SD:
use functions from portage and python-single-r1 instead edo

CC @antecrescent @demize

### DIFF
--- a/games-util/minigalaxy/minigalaxy-1.3.2-r3.ebuild
+++ b/games-util/minigalaxy/minigalaxy-1.3.2-r3.ebuild
@@ -1,0 +1,91 @@
+# Copyright 2020-2025 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+PYTHON_COMPAT=( python3_{11..13} )
+
+inherit desktop optfeature python-single-r1 xdg
+
+DESCRIPTION="A simple GOG client for Linux"
+HOMEPAGE="https://github.com/sharkwouter/minigalaxy"
+SRC_URI="https://github.com/sharkwouter/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+
+IUSE="test"
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
+RESTRICT="!test? ( test )"
+
+# x11-libs/gdk-pixbuf[jpeg] dependency for thumbnails in library entries
+# x11-libs/libnotify dependency in minigalaxy/ui/gtk.py
+RDEPEND="
+	${PYTHON_DEPS}
+	$(python_gen_cond_dep '
+		>=dev-python/requests-2.0.0[${PYTHON_USEDEP}]
+		dev-python/pygobject:3[${PYTHON_USEDEP}]
+	')
+	app-arch/unzip
+	>=net-libs/webkit-gtk-2.6:4.1[introspection]
+	>=x11-libs/gtk+-3[introspection]
+	x11-libs/gdk-pixbuf[introspection,jpeg]
+	x11-libs/libnotify[introspection]
+	x11-misc/xdg-utils
+"
+BDEPEND="
+	${PYTHON_DEPS}
+	sys-apps/help2man
+	sys-devel/gettext
+	test? (
+		$(python_gen_cond_dep '
+			dev-python/simplejson[${PYTHON_USEDEP}]
+		')
+	)
+"
+
+src_test() {
+	epytest
+}
+
+src_install() {
+	sed -e "s:os.path.dirname(sys.argv\[0\]):'${EPREFIX}/usr/share/':" \
+		-e "s:minigalaxy/translations:locale:" \
+		-i minigalaxy/paths.py || die
+
+	mkdir -p usr/share/{minigalaxy,metainfo} || die
+	cp -R data/images data/ui data/style.css usr/share/minigalaxy || die
+	cp data/io.github.sharkwouter.Minigalaxy.metainfo.xml usr/share/metainfo || die
+	insinto /usr
+	doins -r usr/share
+
+	help2man -N -s 6 -n "a simple GTK based GOG Linux client" bin/minigalaxy > minigalaxy.6 || die
+	doman minigalaxy.6
+
+	local lo
+	for lo in data/po/*.po; do
+		msgfmt "${lo}" -o "${lo%.po}.mo" && domo "${lo%.po}".mo || die
+	done
+
+	local x
+	for x in 128 192; do
+		doicon -s ${x} data/icons/${x}x${x}/io.github.sharkwouter.Minigalaxy.png
+	done
+
+	domenu data/io.github.sharkwouter.Minigalaxy.desktop
+
+	dodoc README.md CHANGELOG.md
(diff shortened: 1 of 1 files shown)


PPN-SD:
I just want someone to confirm that it works because I don't use gog

games-util/minigalaxy/minigalaxy-1.3.2-r3.ebuild
```diff
@@ -0,0 +1,91 @@
...
+	sys-devel/gettext
+	test? (
+		$(python_gen_cond_dep '
+			dev-python/simplejson[${PYTHON_USEDEP}]
+		')
+	)
+"
+
+src_test() {
+	epytest
```
gonsoos: Let's use `eunittest test` here. The tests don't depend on pytest and it saves us a dependency.

games-util/minigalaxy/minigalaxy-1.3.2-r3.ebuild
```diff
@@ -0,0 +1,91 @@
...
+		-e "s:minigalaxy/translations:locale:" \
+		-i minigalaxy/paths.py || die
+
+	mkdir -p usr/share/{minigalaxy,metainfo} || die
+	cp -R data/images data/ui data/style.css usr/share/minigalaxy || die
+	cp data/io.github.sharkwouter.Minigalaxy.metainfo.xml usr/share/metainfo || die
+	insinto /usr
+	doins -r usr/share
+
+	help2man -N -s 6 -n "a simple GTK based GOG Linux client" bin/minigalaxy > minigalaxy.6 || die
```
gonsoos: typo: GTK-based

games-util/minigalaxy/minigalaxy-1.3.2-r3.ebuild
```diff
@@ -0,0 +1,91 @@
...
+	doman minigalaxy.6
+
+	local lo
+	for lo in data/po/*.po; do
+		msgfmt "${lo}" -o "${lo%.po}.mo" && domo "${lo%.po}".mo || die
+	done
```
gonsoos: You can call `domo` once with all files. But why not call `scripts/compile-translations.sh` in `src_compile` and then in `src_install`:

```bash
insinto /usr/share/locale
doins -r data/mo/*
```
?

games-util/minigalaxy/minigalaxy-1.3.2-r3.ebuild
```diff
@@ -0,0 +1,91 @@
...
+"
+
+src_test() {
+	epytest
+}
+
+src_install() {
+	sed -e "s:os.path.dirname(sys.argv\[0\]):'${EPREFIX}/usr/share/':" \
+		-e "s:minigalaxy/translations:locale:" \
+		-i minigalaxy/paths.py || die
```
gonsoos, line 55: This belongs in `src_prepare`, which runs before `src_test`, so it also fixes the `eunittest test` invocation.

games-util/minigalaxy/minigalaxy-1.3.2-r3.ebuild
```diff
@@ -0,0 +1,91 @@
...
+
+	mkdir -p usr/share/{minigalaxy,metainfo} || die
+	cp -R data/images data/ui data/style.css usr/share/minigalaxy || die
+	cp data/io.github.sharkwouter.Minigalaxy.metainfo.xml usr/share/metainfo || die
+	insinto /usr
+	doins -r usr/share
```
gonsoos: We don't need external commands to do this :)
```suggestion
	insinto /usr/share/minigalaxy
	doins -r data/images data/ui data/style.css
	insinto /usr/share/metainfo
	doins data/io.github.sharkwouter.Minigalaxy.metainfo.xml
```

antecrescent (changes requested):
Nice! This is a significant improvement IMO :) Apart from my comments, please change the commit title to something more meaningful, since you effectively change the packaging approach, fix install locations and compile man pages, not just remove `edo`.

> I just want someone to confirm that it works because I don't use gog

Works for me!

games-util/minigalaxy/minigalaxy-1.3.2-r3.ebuild
PPN-SD: pytest is really really nicer no ?

games-util/minigalaxy/minigalaxy-1.3.2-r3.ebuild
PPN-SD: > You can call `domo` once with all files. But why not call `scripts/compile-translations.sh` in `src_compile` and then in `src_install`:
> 
> ```bash
> insinto /usr/share/locale
> doins -r data/mo/*
> ```
> ?

I'll test with one domo invocation, thanks

games-util/minigalaxy/minigalaxy-1.3.2-r3.ebuild
PPN-SD: paths.py as two ways to find files. First try is for src-tree, second for root-tree. Did you get failures ?

games-util/minigalaxy/minigalaxy-1.3.2-r3.ebuild
gonsoos: In what sense?

PPN-SD: output results, nothing significant

games-util/minigalaxy/minigalaxy-1.3.2-r3.ebuild
gonsoos: You can use this in `src_compile`
```bash
	lo_files=( data/po/*.po )
	local lo
	for lo in "${lo_files[@]%.po}"; do
		msgfmt "${lo}.po" -o "${lo}.mo" || die
	done
```
and this in `src_install`
```bash
	domo "${lo_files[@]/%.po/.mo}"
```

PPN-SD: domo data/po/*.po just works as expected

games-util/minigalaxy/minigalaxy-1.3.2-r3.ebuild
PPN-SD: eunittest **tests** defines correct paths

gonsoos: > , so it also fixes the eunittest test invocation.

Yeah right, that was bogus. Sorry, that's on me.

> This belongs in src_prepare, 

Still, source manipulations belong in `src_prepare`. We want to test the package that we're going to install afterwards and not introduce additional changes after testing.

games-util/minigalaxy/minigalaxy-1.3.2-r3.ebuild
gonsoos: > domo data/po/*.po just works as expected

Unfortunately not. This just installs the `.po` files with a `.mo` file extension.
https://gitweb.gentoo.org/proj/portage.git/tree/bin/ebuild-helpers/domo?h=portage-3.0.67#n43

Fixed in 7bae30c96ecf496380378000fabde232b1370b2e

PPN-SD: I misunderstood your comment. Thanks for the fix.

PR #331, opened 2025-06-08
dev-libs/aws-sdk-cpp: add package aws-sdk-cpp

KotoishiHeart:
Add aws-sdk-cpp and its dependencies to guru.

If we add it, it will be after June 19th when the dependency will be removed from the gentoo repository.

lists:
aws-sdk-cpp
aws-c-auth
aws-c-cal
aws-c-common
aws-c-compression
aws-c-event-stream
aws-c-http
aws-c-io
aws-c-mqtt
aws-c-s3
aws-c-sdkutils
aws-checksums
aws-crt-cpp
s2n


antecrescent:
Great effort bumping these packages!
Why not clean up the contributions, open a PR against ::gentoo and maintain them there? GURU is not a dumping ground for unmaintained packages that are treecleaned from ::gentoo.
We do accept packages from the parent repo though, e.g. if it's a dependency of one of our packages.

KotoishiHeart:
Since aws-sdk-cpp is not available in ::gentoo, I personally think it would be better to move it to guru rather than maintaining it on gentoo.
I'm not going to make guru the place to just spit out aws-sdk-cpp and its dependencies.

dev-libs/aws-c-auth/aws-c-auth-0.9.0.ebuild
```diff
@@ -0,0 +1,34 @@
...
+SRC_URI="https://github.com/awslabs/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="Apache-2.0"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="static-libs test"
+
+RESTRICT="!test? ( test )"
+
+DEPEND="
```
thesamesam: If USE=-static-libs, surely these are all RDEPENDs too? Honestly, we usually don't worry about making the static-libs deps purely correct (as in: making sure to avoid anything in RDEPEND in the USE=static-libs case). Besides, USE=static-libs usually means "static libs as well", not "only static libs".

thesamesam: But do we really need this at all? Can't we rip out all this static-libs stuff? By policy, we avoid it wherever possible.

dev-libs/aws-sdk-cpp/aws-sdk-cpp-1.11.586.ebuild
```diff
@@ -0,0 +1,100 @@
...
+			mybuildtargets+=";$( _aws_expand_group "${g}" )"
+		fi
+	done
+
+	local mycmakeargs=(
+		-DAUTORUN_UNIT_TESTS=$(usex test)
+		-DAWS_SDK_WARNINGS_ARE_ERRORS=OFF
+		-DBUILD_DEPS=NO
+		-DBUILD_ONLY="${mybuildtargets}"
+		-DBUILD_SHARED_LIBS=ON
```
thesamesam: The eclass sets this for us.

dev-libs/aws-sdk-cpp/aws-sdk-cpp-1.11.586.ebuild
```diff
@@ -0,0 +1,100 @@
...
+		-DCPP_STANDARD=17
+		-DENABLE_RTTI=$(usex rtti)
+		-DENABLE_TESTING=$(usex test)
+		-DENABLE_UNITY_BUILD=$(usex unity-build)
```
thesamesam: We usually use `jumbo-build` for this but let's please not make it a USE flag unless there's a really compelling reason.

dev-libs/aws-sdk-cpp/aws-sdk-cpp-1.11.586.ebuild
```diff
@@ -0,0 +1,100 @@
...
+	done
+
+	local mycmakeargs=(
+		-DAUTORUN_UNIT_TESTS=$(usex test)
+		-DAWS_SDK_WARNINGS_ARE_ERRORS=OFF
+		-DBUILD_DEPS=NO
+		-DBUILD_ONLY="${mybuildtargets}"
+		-DBUILD_SHARED_LIBS=ON
+		-DCPP_STANDARD=17
+		-DENABLE_RTTI=$(usex rtti)
```
thesamesam: Does this really need to be a USE flag?

dev-libs/s2n/s2n-1.5.21.ebuild
```diff
@@ -0,0 +1,46 @@
...
+
+S="${WORKDIR}/${PN}-tls-${PV}"
+
+src_prepare() {
+	cmake_src_prepare
+
+	# Fix shared library building, needed for USE="test"
+	# See: https://github.com/awslabs/s2n/issues/2401
+	if use test; then
+		sed -i -e 's, -fvisibility=hidden,,' "${S}"/CMakeLists.txt || die "sed failed"
```
thesamesam: Changing how the rest of the source code is built depending on if tests are enabled means we're not testing what we're deploying most of the time.

thesamesam (changes requested):

dev-libs/s2n/s2n-1.5.21.ebuild
KotoishiHeart: After checking, I found that this issue has been fixed and there is no need to modify CMakeLists.txt, so I will delete it.

dev-libs/aws-sdk-cpp/aws-sdk-cpp-1.11.586.ebuild
KotoishiHeart: Does this mean that I need to use ‘usex jumbo-build’ instead of ‘usex unity-build’?

dev-libs/aws-c-auth/aws-c-auth-0.9.0.ebuild
KotoishiHeart: I think it's okay without static-libs, so I'll delete it.

dev-libs/aws-sdk-cpp/aws-sdk-cpp-1.11.586.ebuild
thesamesam: Yes, but if possible, it is better to drop it entirely and just disable it, as it harms ccache/debugging if things go wrong/consumes more resources.

KotoishiHeart: I understand. We'll remove useflag since there's no particular reason to need it.

dev-libs/aws-sdk-cpp/aws-sdk-cpp-1.11.586.ebuild
KotoishiHeart: BUILD_SHARED_LIBS was ON by default.
Since we are going to delete static-libs, there is no reason to keep it, so we will delete it.

What items need to be specified in an eclass?

thesamesam: cmake.eclass sets `-DBUILD_SHARED_LIBS=ON` for us so we don't need to set it. But I don't think you have any other redundant options here.

KotoishiHeart: I understand. thank you.

dev-libs/aws-sdk-cpp/aws-sdk-cpp-1.11.586.ebuild
KotoishiHeart: RTTI is enabled by default, so this flag could be removed. i delete.

KotoishiHeart:
I have revised the content as pointed out.

dev-libs/aws-c-compression/aws-c-compression-0.3.1.ebuild
```diff
@@ -0,0 +1,28 @@
+# Copyright 1999-2022 Gentoo Authors
```
gonsoos: This is outdated.

dev-libs/aws-checksums/aws-checksums-0.2.7.ebuild
```diff
@@ -0,0 +1,28 @@
+# Copyright 1999-2022 Gentoo Authors
```
gonsoos: This is outdated.

dev-libs/aws-c-event-stream/aws-c-event-stream-0.5.4.ebuild
```diff
@@ -0,0 +1,40 @@
...
+SRC_URI="https://github.com/awslabs/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="Apache-2.0"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="test"
+
+RESTRICT="!test? ( test )"
+
+DEPEND="
```
thesamesam: RDEPEND is empty? Does it not link with any of these libraries?

(Please check this everywhere.)

dev-libs/aws-c-cal/aws-c-cal-0.9.2.ebuild
```diff
@@ -0,0 +1,29 @@
...
+
+LICENSE="Apache-2.0"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE="test"
+
+RESTRICT="!test? ( test )"
+
+DEPEND="
+	dev-libs/aws-c-common:0=
```
thesamesam: Please use just `:=` unless there's a realistic prospect of these being slotted in future (unlikely IMO).

dev-libs/aws-c-auth/aws-c-auth-0.9.0.ebuild
```diff
@@ -0,0 +1,33 @@
+# Copyright 1999-2025 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit cmake
+
+DESCRIPTION="C99 library implementation of AWS client-side authentication: standard credentials providers and signing."
```
thesamesam: No fullstops at the end of DESCRIPTION:
```suggestion
DESCRIPTION="C99 library implementation of AWS client-side authentication: standard credentials providers and signing"
```

KotoishiHeart:
fixed ebuilds

dev-libs/aws-sdk-cpp/aws-sdk-cpp-1.11.586.ebuild
```diff
@@ -0,0 +1,97 @@
...
+DESCRIPTION="AWS SDK for C++"
+HOMEPAGE="https://aws.amazon.com/sdk-for-cpp/"
+SRC_URI="https://github.com/aws/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="Apache-2.0"
+SLOT="0"
+KEYWORDS="~amd64"
+
+# BUILD_ONLY lists
+AWS_GROUP_storage="backup;backup-gateway;datasync;ebs;fsx;glacier;mediapackage;mediapackage-vod;mediapackagev2;mediastore;mediastore-data;s3;s3-crt;snow-device-management;snowball;storagegateway;workdocs;awstransfer;importexport;rbin;s3tables"
```
thesamesam: I'm undecided on if I'm okay with this, it seems kind of wild and wonder if we should be enabling all of these by default or something if none of them have extra deps and if they don't have signifiant build-time impact (perhaps they do though -- if so, say it in the commit message).

But supposing I am okay with this: don't we need USE descriptions for these in metadata.xml? pkgcheck should warn on that? Or am I missing something?

KotoishiHeart: They do not add any new dependencies. However, they are too large to be enabled all the time, so they are not enabled by default except for the minimum core, identity-management, and sts.

It is so huge that even compiling with 16 threads using a Ryzen 9 7900 would take about 30 minutes.

I would say so in the metadata.xml and in the commit message.

thesamesam: > I would say so in the metadata.xml and in the commit message.

OK, let's give it a go, but I reserve the right to request tweaks later. Feel free to merge with that change.

Thanks for the good work and responses to comments.

KotoishiHeart: Added. Please let us know if we are missing any information you need.

thesamesam: Commit message should say why we have so many (extensive build time, quote the measurement you gave me) then lgtm

KotoishiHeart: i understand. maybe fixed.

KotoishiHeart:
Thank you for confirming.

antecrescent:
@KotoishiHeart Could you also take care of these issues, please?

Most importantly the DeprecatedDep, missing metadata.xml and UnknownUseFlags.
Thanks!
```
 $ pkgcheck scan --commits e3a9b41dfc95e30e9f579bb621c6c475fc98cf7a..HEAD
dev-libs/aws-c-event-stream
  MissingRemoteId: missing <remote-id type="github">awslabs/aws-c-event-stream</remote-id> (inferred from URI 'https://github.com/awslabs/aws-c-event-stream')

dev-libs/aws-checksums
  MissingRemoteId: missing <remote-id type="github">awslabs/aws-checksums</remote-id> (inferred from URI 'https://github.com/awslabs/aws-checksums')

dev-libs/aws-c-auth
  MissingRemoteId: missing <remote-id type="github">awslabs/aws-c-auth</remote-id> (inferred from URI 'https://github.com/awslabs/aws-c-auth')
  BadDescription: version 0.9.0: over 80 chars in length

dev-libs/aws-c-compression
  MissingRemoteId: missing <remote-id type="github">awslabs/aws-c-compression</remote-id> (inferred from URI 'https://github.com
/awslabs/aws-c-compression')

dev-libs/s2n
  VariableOrderWrong: version 1.5.21: variable S should occur before RESTRICT

dev-libs/aws-sdk-cpp
  DeprecatedDep: version 1.11.586: DEPEND: deprecated dependency: media-sound/pulseaudio
  DeprecatedDep: version 1.11.586: RDEPEND: deprecated dependency: media-sound/pulseaudio
  ExcessiveLineLength: version 1.11.586: excessive line length (over 120 characters) on line: 34
  PythonCompatUpdate: version 1.11.586: PYTHON_COMPAT update available: python3_14
  UnknownUseFlags: version 1.11.586: unknown USE flags: 'full', 'http'

```

antecrescent:
> I don't know how to resolve DeprecatedDep.

I believe the reason you include the package is the following bit in `src/aws-cpp-sdk-text-to-speech/CMakeLists.txt`, correct?
```
check_include_files("pulse/simple.h" HAVE_PULSE)
```
The dependency that provides this header is `media-libs/libpulse`. That CMakeLists.txt file needs to be fixed as well, because it adds a dependency on libpulse, regardless of the `pulseaudio` USE flag setting, as long as the header is available on the system.

Since text-to-speech support seems to strongly depend on libpulse, I'd probably replace the `pulseaudio` USE flag with `speech` and build the text-to-speech SDK conditional on that or build it unconditionally and remove the USE flag.

KotoishiHeart:
> Since text-to-speech support seems to strongly depend on libpulse, I'd probably replace the pulseaudio USE flag with speech and build the text-to-speech SDK conditional on that or build it unconditionally and remove the USE flag.

I want to enable "speech", but do I need to add a DEPEND on "media-libs/libpulse"?

KotoishiHeart:
fixed

PR #334, opened 2025-06-16
dev-python/internetarchive python 3.13 add and add 5.3.0

vitaly-zdanevich:


dev-python/internetarchive/Manifest
```diff
@@ -1 +1,2 @@
 DIST v3.7.0.tar.gz 145990 BLAKE2B f4832193fde0afdfd6ea1af46d0dea08abd73e56b5d799eb4e9a1b6424f95dbb6b1d0bc3d25fe0ae2bda7edcd003cfbff97c941a313018d4fb83bb562e30c121 SHA512 56554c3001f21ce770fd44ed2628b2fac40b96bcd5370935604c1fec465fd5a44be4a89ffa119cbd6f01be805161fe2fac93375d96f59d2ed38ff6418d5da714
+DIST v5.3.0.tar.gz 152973 BLAKE2B 1ff3b3bb61d830a27738bded36346c599bd9adbe34c4627a88ed8a371249701c7219e798211e82f58a4e4207e83761154c1b43b4046d3cfbaeb2d0dfd414ad3f SHA512 8f0e07e5add97a8f4d194fef023388c702c17bdd1bf25b8b22e7962a53f2a8077577eb615d8eb74769a829bac5ec41aeb5617ea4fc8e38d4adb0be69b2a896f3
```
thesamesam: The distfile name here is too generic and must be renamed.

dev-python/internetarchive/internetarchive-5.3.0.ebuild
```diff
@@ -0,0 +1,31 @@
...
+
+DESCRIPTION="Lib and CLI for archive.org - for search, uploading, downloading, rename..."
+HOMEPAGE="https://github.com/jjjake/internetarchive"
+SRC_URI="https://github.com/jjjake/$PN/archive/refs/tags/v$PV.tar.gz"
+
+LICENSE="AGPL-3"
+SLOT="0"
+KEYWORDS="~amd64"
+
+RESTRICT="test"
```
thesamesam: Why aren't tests wired up?

dev-python/internetarchive/internetarchive-5.3.0.ebuild
```diff
@@ -0,0 +1,31 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+PYTHON_COMPAT=( python3_{11..13} )
+DISTUTILS_USE_PEP517=setuptools
+inherit distutils-r1
+
+DESCRIPTION="Lib and CLI for archive.org - for search, uploading, downloading, rename..."
+HOMEPAGE="https://github.com/jjjake/internetarchive"
+SRC_URI="https://github.com/jjjake/$PN/archive/refs/tags/v$PV.tar.gz"
```
thesamesam: Gentoo style is `${PN}` and `${PV}`.

dev-python/internetarchive/internetarchive-5.3.0.ebuild
```diff
@@ -0,0 +1,31 @@
...
+	dev-python/docopt[${PYTHON_USEDEP}]
+	dev-python/jsonpatch[${PYTHON_USEDEP}]
+	dev-python/requests[${PYTHON_USEDEP}]
+	dev-python/schema[${PYTHON_USEDEP}]
+	dev-python/tqdm[${PYTHON_USEDEP}]
+	dev-python/urllib3[${PYTHON_USEDEP}]
+"
+
+pkg_postinst() {
+	einfo "See documentation at https://archive.org/developers/internetarchive/"
```
thesamesam: Not sure this is really needed. Just put it in HOMEPAGE?

vitaly-zdanevich: But with HOMEPAGE user will not see it after install :(

stkw0: If users want to see the homepage, there is a variable for the HOMEPAGE and proper tools to get that information. 
Most users will probably just ignore it, and when they needed, they will use the standard tools. To have it in pkg_postinst is harmless but pointless.

dev-python/internetarchive/internetarchive-3.7.0.ebuild
```diff
@@ -1,20 +1,27 @@
...
 EAPI=8
-PYTHON_COMPAT=( python3_{11..12} )
+PYTHON_COMPAT=( python3_{11..13} )
 DISTUTILS_USE_PEP517=setuptools
 inherit distutils-r1
 
 DESCRIPTION="Lib and CLI for archive.org - for search, uploading, downloading, rename..."
 HOMEPAGE="https://github.com/jjjake/internetarchive"
-SRC_URI="https://github.com/jjjake/$PN/archive/refs/tags/v$PV.tar.gz"
+SRC_URI="https://github.com/jjjake/${PN}/archive/refs/tags/v${PV}.tar.gz -> ${P}.tar.gz"
```
stkw0: To make it even better it should be ${P}.gh.tar.gz as the source code can also be grabbed from https://pypi.org/project/internetarchive/

dev-python/internetarchive/internetarchive-3.7.0.ebuild
```diff
@@ -1,20 +1,27 @@
...
 
 LICENSE="AGPL-3"
 SLOT="0"
 KEYWORDS="~amd64"
 
-RESTRICT="test"
+BDEPEND="
+	test? (
+		dev-python/pytest[${PYTHON_USEDEP}]
```
stkw0: You don't need this

dev-python/internetarchive/internetarchive-5.3.0.ebuild
gonsoos: IMO, this belongs metadata.xml with a `doc` tag.

dev-python/internetarchive/internetarchive-3.7.0.ebuild
```diff
@@ -1,20 +1,26 @@
-# Copyright 2024 Gentoo Authors
+# Copyright 2025 Gentoo Authors
```
gonsoos: 2024-2025

dev-python/internetarchive/internetarchive-5.3.0.ebuild
```diff
@@ -0,0 +1,39 @@
+# Copyright 2025 Gentoo Authors
```
gonsoos: 2024-2025

dev-python/internetarchive/internetarchive-5.3.0.ebuild
```diff
@@ -0,0 +1,39 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+PYTHON_COMPAT=( python3_{11..13} )
+DISTUTILS_USE_PEP517=setuptools
+inherit distutils-r1
+
+DESCRIPTION="Lib and CLI for archive.org - for search, uploading, downloading, rename..."
```
gonsoos: No ellipses or full stops at the end of DESCRIPTION. The upstream description reads:
> A Python and Command-Line Interface to Archive.org

dev-python/internetarchive/internetarchive-5.3.0.ebuild
```diff
@@ -0,0 +1,39 @@
...
+HOMEPAGE="https://github.com/jjjake/internetarchive"
+SRC_URI="https://github.com/jjjake/${PN}/archive/refs/tags/v${PV}.tar.gz -> ${P}.gh.tar.gz"
+
+LICENSE="AGPL-3"
+SLOT="0"
+KEYWORDS="~amd64"
+
+RESTRICT="test"
```
gonsoos: Now that you're wiring up tests, this needs to be removed so that tests actually run, if the user has `FEATURES=test`.

dev-python/internetarchive/internetarchive-5.3.0.ebuild
thesamesam: FWIW, we often avoid "A.."

vitaly-zdanevich:
Hm, I tried `USE=test emerge -a dev-python/internetarchive` and see
`>>> Test phase [not enabled]: dev-python/internetarchive-3.7.0`

vitaly-zdanevich:
@jjjake hi, what is the correct way to disable networked tests? Or should you add mocks? Because on Gentoo, during installation - the network is not available.

pHlt7:
> @jjjake hi, 

Please use the appropriate help channels, such as `#gentoo-dev-help` on IRC, when encountering issues during ebuild development instead of involving upstream.

The test failures are a Gentoo-specific issue, because we build and test packages with `FEATURES=network-sandbox`, which removes network access during the ebuild phase functions as the name implies. > what is the correct way to disable networked tests? Or should you add mocks? Because on Gentoo, during installation - the network is not available.

> what is the correct way to disable networked tests?

Try disabling the networked tests using the `EPYTEST_DESELECT` array. You can reference Gentoo's Python Guide for how to use it:
https://projects.gentoo.org/python/guide/pytest.html#skipping-tests-based-on-paths-names

pHlt7:
Please split the commits like I asked you to and add a Signed-off-by tag.

pHlt7:
The first commit message is still incorrect. Also, dependencies must be defined before calling `distutils_enable_tests`. In order: `RDEPEND`, then `BDEPEND`, then `EPYTEST_DESELECT` and lastly `distutils_enable_tests`.

vitaly-zdanevich:
> dependencies must be defined before calling

Should `pkgcheck scan` say that?

pHlt7:
>> dependencies must be defined before calling

> Should pkgcheck scan say that?

It could :shrug: 



>> In order: RDEPEND, then BDEPEND

> But more logical Build before Run?

While uncommon, in hindsight that's also acceptable. I thought `pkgcheck scan` enforces the ordering of the *DEPEND variables in [`skel.ebuild`](https://github.com/gentoo/gentoo/blob/master/skel.ebuild) but it doesn't and for good reason.

PR #335, opened 2025-06-16
app-text/writedoc: new package for fast file creation and organisation

Chiron8:
This pr adds app-text/writedoc

This is a simple utility that helps create files without needing to define file paths or types for the purpose of quick note taking. It supports changing default directories and file types.

Written in Bash
License - MIT
Tested on ~amd64 and ~x86

Remote-ID: [GitHub - Chiron8/writedoc](https://github.com/Chiron8/writedoc)

Signed-off-by: Chiron Horwood chiron.dev8@gmail.com


app-text/writedoc/writedoc-2.2.ebuild
```diff
@@ -0,0 +1,28 @@
+EAPI=8
```
thesamesam, line 1: Please always run `pkgcheck scan` and `pkgcheck scan --commits`.

app-text/writedoc/writedoc-2.2.ebuild
```diff
@@ -0,0 +1,28 @@
...
+DESCRIPTION="Quick file creation without file paths"
```
thesamesam, line 2: The ebuild is missing the copyright header and the formatting is also strange, missing whitespace/blank lines everywhere.

app-text/writedoc/writedoc-2.2.ebuild
```diff
@@ -0,0 +1,28 @@
...
+HOMEPAGE="https://github.com/Chiron8/writedoc"
+SRC_URI="https://github.com/Chiron8/writedoc/archive/refs/tags/v${PV}.tar.gz -> ${P}.tar.gz"
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+FEATURES="test"
```
thesamesam, line 8: This is not something ebuilds should set.

app-text/writedoc/writedoc-2.2.ebuild
```diff
@@ -0,0 +1,28 @@
...
+src_test() {
+nvim -v >> /dev/null || die "neovim not installed"
+}
+src_install() {
+install -Dm755 writedoc "${D}/usr/bin/writedoc" || die
+install -Dm644 writedoc.1 "${D}/usr/share/man/man1/writedoc.1" || die
+dodoc README.md
+}
+pkg_postinst() {
+ewarn "Running post install tests"
```
thesamesam, line 22: We do not have a facility for post-merge testing. We have some plans for it, but it's only necessary for some special cases where privileges are needed and such, not this.

The code here should really be in `src_test` instead. Also, in `pkg_postinst`, this would break for cross-compilation.

app-text/writedoc/writedoc-2.2.ebuild
```diff
@@ -0,0 +1,28 @@
...
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+FEATURES="test"
+RDEPEND="
+app-shells/bash
+app-editors/neovim
+"
+src_test() {
+nvim -v >> /dev/null || die "neovim not installed"
```
thesamesam, line 14: It seems impossible for this to fail given neovim is a dependency, but it's also not testing writedoc itself?

app-text/writedoc/writedoc-2.2.ebuild
```diff
@@ -0,0 +1,28 @@
...
+}
+src_install() {
+install -Dm755 writedoc "${D}/usr/bin/writedoc" || die
```
thesamesam, line 17: * Is there no proper build system upstream?
* Please use `doins` and friends if you must.
* Supposing you must use `install`, you should have `ED` in there.

Chiron8:
Thank you for the feedback. I'll keep this local for now.

PR #298, opened 2025-03-03
dev-cpp/fast_obj: new package

vitaly-zdanevich:

PR #297, opened 2025-03-01
gui-apps/organicmaps: fix compilation

vitaly-zdanevich:
In the next PR will add versioned ebuild.

PR #336, opened 2025-06-17
new package

vitaly-zdanevich:
From https://github.com/onkelbeh/HomeAssistantRepository/tree/master/dev-python/pyicloud


stkw0:
It's not acceptable to just make a copy-paste and expect it would be merged. You can close this PR, or look for yourself what is wrong and fix it.

vitaly-zdanevich:
Fixed dependencies.

stkw0:
Starting with the most formal things:
 - commit don't have the Signed-off-by tag
 - The change to the DESCRIPTION makes it more obscure to understand what is this package about
 - you just copy-pasted a pretty old version, why you didn't even cared to update it?
 - we only store the hashes for DIST entries, not EBUILD or MISC. Metadata file should be regenerated
 - Yes, the maintainer file was wrong, it is still wrong as it should have the maintainer needed comment.
 - As you might have noticed, the minimum python version was being moved to pyhon3_11
 - you clearly didn't run pkgcheck, which is supposed to be a tool to help find common mistakes
 - KEYWORDS should only list the arch that has been tested. And you probably didn't test the arm ones
- (maybe something else I miss)

Most of it, individually, could be taken as a small slip. However, in this case it was clearly not a mistake but a total carelessness. Even worst given that you have been around for ¿years? so you were probably aware of the most basic things (so it's not because of ignorance). As a result, I don't think you are competent enough and that it would be bad for the maintenance of the repository to add your ebuilds or the ones you send, unless there is a more capable person that is willing to take care of maintaining them. 

I would not review or merge your PR that adds a new package from now on, and it's my opinion that you should avoid sending PR for new packages (and close the ones that add new packages).

vitaly-zdanevich:
Fixed everything.

stkw0:
Last version is the 2.0.1 from 4 days ago. See https://pypi.org/project/pyicloud/#history, https://github.com/timlaing/pyicloud and github.com/pypi/support/issues/5377

vitaly-zdanevich:
> close the ones that add new packages

https://github.com/gentoo/guru/pull/298 is needed to fix Organicmaps https://github.com/gentoo/guru/pull/297

MrRoy:
> Replaced the upstream.

Not in metadata.xml

PR #337, opened 2025-06-22
WIP: Drop Qt5 support

gonsoos:
::gentoo is actively dropping Qt5 support, we should too. (https://bugs.gentoo.org/948836)

 I try to continuously update this PR and ask the respective maintainers for reviews (and runtime-testing ideally).
@maharmstone: `app-benchmarks/kdiskmark`
@kostadinsh : `app-misc/corectrl`

Outstanding Qt5-ebuilds that I'm working are here:
https://github.com/antecrescent/guru/tree/qt5-investigate


kostadinsh:
No objections from me for my package.

app-arch/libarchive-qt/libarchive-qt-2.0.8-r1.ebuild
```diff
@@ -13,20 +13,20 @@ S="${WORKDIR}/${PN}-v${PV}"
 LICENSE="LGPL-3+"
 SLOT="0"
 KEYWORDS="~amd64"
-IUSE="qt6"
```
pastalian, line 16: Can you also change
`app-arch/libarchive-qt[qt6]` to either
`app-arch/libarchive-qt[qt6(+)]` or just `app-arch/libarchive-qt`
in `gui-apps/corearchiver`

gonsoos: Oh, thanks! Does the change `[qt6]` -> `[qt6(+)]` warrant a new revision?

pastalian: Yeah, I think so.

gonsoos: After discussing this on IRC, I've come to the conclusion that it's not necessary to revbump here. Portage seems to factor in the new USE default without a rebuild of corearchive, too, so it's still able to update libarchive-qt.

eli-schwartz: It is necessary.

PR #338, opened 2025-06-25
dev-cpp/wt: revert adding qtbase dep

PPN-SD:
revert 8a6c242
wt needs qt only for building [examples](https://github.com/search?q=repo%3Aemweb%2Fwt+%2F(%3F-i)include+<Q%2F) that are already disabled

@stkw0 @antecrescent


antecrescent (approved):
You are correct. I didn't stop to check, what part of the codebase actually depends on Qt and got misled by 
```
IF(ENABLE_QT6)
  FIND_PACKAGE(Qt6 CONFIG COMPONENTS Core)
ENDIF(ENABLE_QT6)
```
in CMakeLists.txt.

Sorry, that's totally on me!
I'm waiting for @stkw0 to confirm this.

PR #339, opened 2025-06-27
dev-cpp/wt: rework deps, unrestrict boost, remove old

PPN-SD:
@stkw0 After https://github.com/gentoo/guru/commit/60dc0de5a4c63c8493852510d6dc61f4a99a0df6, I tried to polish the ebuild.

dep :
unrestrict boost version. It builds now with [>=1.87](https://github.com/emweb/wt/commit/f39d3a00e0b3f9bdda60b061c80001420f6e2703)
media-libs/libharu is only required with pdf feature
add media-libs/glew required for opengl
add connector-c for mariadb/mysql
add pkgconfig
use libglvnd[X] instead of virtual/opengl

doc :
generate qch requires qttools[assistant] not qttools[qdoc]
add dev-ruby/asciidoctor for tutorial building
add graphviz for dot

add useflags :
make graphicsmagick (raster image) optional
make Wt::Test optional with wttest
make sqlite optional (but enabled by default) and force system-lib

REQUIRED_USE :
pango is only used with graphicsmagick or libharu
opengl requires graphicsmagick


dev-cpp/wt/wt-4.11.4-r3.ebuild
```diff
@@ -0,0 +1,126 @@
...
+	)
+	pdf? ( media-libs/libharu:= )
+	postgres? ( dev-db/postgresql )
+	sqlite? ( dev-db/sqlite:3 )
+	ssl? ( dev-libs/openssl:= )
+"
+RDEPEND="${DEPEND}"
+BDEPEND="
+	doc? (
+		app-text/doxygen[dot]
```
gonsoos: Replace this with
```
app-text/doxygen
 media-gfx/graphviz[use1,use2,...]
```
 See https://bugs.gentoo.org/430914.

dev-cpp/wt/wt-4.11.4-r3.ebuild
```diff
@@ -0,0 +1,126 @@
...
+	mysql? (
+		virtual/mysql
+		|| (
+			dev-db/mariadb-connector-c
+			dev-db/mysql-connector-c
+		)
+	)
+	opengl? (
+		media-libs/glew:=
+		virtual/opengl
```
gonsoos: ::gentoo has somewhat started to deprecate `virtual/opengl` in favor of the actual libraries that the package links against: https://wiki.gentoo.org/wiki/User:ManDay/Deprecating_the_virtual_OpenGL_ebuild

Please list these instead.

dev-cpp/wt/wt-4.11.4-r3.ebuild
```diff
@@ -0,0 +1,126 @@
...
+	ssl? ( dev-libs/openssl:= )
+"
+RDEPEND="${DEPEND}"
+BDEPEND="
+	doc? (
+		app-text/doxygen[dot]
+		dev-qt/qttools:6[assistant]
+		dev-ruby/asciidoctor
+	)
+"
```
gonsoos, line 57: Does the build system successfully pick up pango, when `virtual/pkgconfig` is missing? If not, you may be missing a `pango? ( virtual/pkgconfig )` dependency. I'm not sure, if one can test this without breaking `dev-build/cmake`, which depends on pkgconfig at runtime.

See:
https://github.com/emweb/wt/blob/c158c796dd10f44fdd0fce60b9a2f4215927347a/cmake/WtFindPangoFt2.txt#L1
pulled in by
https://github.com/emweb/wt/blob/c158c796dd10f44fdd0fce60b9a2f4215927347a/CMakeLists.txt#L278

antecrescent (changes requested):
How likely is one to want to render raster images? If this is a very common use case, I'd consider forcing `-DWT_WRASTERIMAGE_IMPLEMENTATION=GraphicsMagick`, which allows us to remove `REQUIRED_USE` altogether.

This is just a suggestion though.

dev-cpp/wt/wt-4.11.4-r3.ebuild
PPN-SD: Right. Add graphviz as dot tool is required.

dev-cpp/wt/wt-4.11.4-r3.ebuild
PPN-SD: added unconditionally.

dev-cpp/wt/wt-4.11.4-r3.ebuild
PPN-SD: media-libs/glew is the old way too but it's required. I leave as-is now.

stkw0:
> > How likely is one to want to render raster images? If this is a very common use case, I'd consider forcing `-DWT_WRASTERIMAGE_IMPLEMENTATION=GraphicsMagick`, which allows us to remove `REQUIRED_USE` altogether.
> > This is just a suggestion though.
> 
> I use dev-cpp/wt for fileshelter::guru and render raster images is not needed.

 I still didn't looked at the PR properly, thank you very much for the various fixes.
In regards to the USE flags, even if it's not needed for a particular app, the cost of enabling it are negligible and the deps are not big, that's why I left sqlite and graphicsmagick  unconditionally enabled and if there is no use case that needs them to be disabled I think it's better to let it enabled.

dev-cpp/wt/wt-4.11.4-r3.ebuild
gonsoos: > media-libs/glew is the old way too but it's required.

I don't understand "is the old way". Do you mean that the library depends on `virtual/opengl` too? In any case, us depending on `media-libs/glew` is correct so let's keep it.

What I suggest we do is replace `virtual/opengl` with `media-libs/libglvnd[X]` because of
```
$ scanelf -n   usr/lib64/libwt.so.4.11.4
 TYPE   NEEDED FILE
ET_DYN [...] ,libGLX.so.0,libOpenGL.so.0, [...] usr/lib64/libwt.so.4.11.4
```
as suggested in the link above (and by extension in https://bugs.gentoo.org/890334#c5).

PPN-SD: I mean media-libs/glew itself depends on virtual/opengl. But no matter. Changed.

stkw0:
> By default, it's a quite old bundled sqlite 3.35.5, with a lot of CVE since I guess.

We are not using the bundled sqlite (or should not if there is one)¿?

About the other, I really appreciate the fix. 

I will not have time too look this into detail maybe until Monday.  But at first glance, the only other think makes me a bit dubious is the wttest USE, why not just have test use flag?

stkw0:
> > We are not using the bundled sqlite  (or should not if there is one)¿?
> 
> The current ebuild uses bundled sqlite. Sorry I don’t understand the question/comment.

That if now we are forcing always the system sqlite instead of the bundled one, it will not be the old vulnerable one.

dev-cpp/wt/wt-4.11.4-r3.ebuild
```diff
@@ -0,0 +1,131 @@
...
+
+src_prepare() {
+	cmake_src_prepare
+
+	# remove bundled sqlite
+	rm -r src/Wt/Dbo/backend/amalgamation || die
+
+	if use doc; then
+		doxygen -u Doxyfile 2>/dev/null || die
+		doxygen -u examples/Doxyfile 2>/dev/null || die
```
stkw0, line 69: Is it worthy to install the docs of the examples given that examples are unconditionally disabled?

PPN-SD: This is not the same. It's not doc for wtwithqt.

stkw0:
I just merged it. Thank you very much for the number of fixes. At least I would have liked to remove the wttest use and default to ON as it's a single file, but unfortunately it would break the LTO builds.

btw, feel free to add yourself as a maintainer if you want.

PR #340, opened 2025-06-29
www-apps/redlib: small fixes and update to 0.36.0

tokoyami:
I dropped the old versions as they are effectively no longer functional.


pHlt7:
@MrRoy Pinging the maintainer of the package for a review in case they haven't seen this PR.

stkw0:
I think it would be better to fixed up the 3 commits related with env vars in a single commit

PR #343, opened 2025-07-08
TODO.md: remove croissant

sfarajli:
dev-lua/croissant is now present, so remove it from the todo list.


pHlt7:
Merged, thank you!

(This was missing a "Signed-off-by" line, which is OK because a one-line
change isn't copyrightable. However, please add that line for your next
contribution. See GLEP 76 [1] for our copyright policy.)

[1] https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin

PR #345, opened 2025-07-10
dev-libs/hardened_malloc: new package, add 13

tmak2002:


tmak2002:
> savedconfig is probably more appropriate here.

Sry for the late answer. Can you explain why savedcondig better is for this? I dont see a big benefit over the actual approach.

dev-libs/hardened_malloc/hardened_malloc-13.ebuild
```diff
@@ -0,0 +1,29 @@
...
+      emake VARIANT=default
+  fi
+}
+
+src_install() {
+  if use light; then
+    dolib.so out-light/libhardened_malloc-light.so
+  else
+    dolib.so out/libhardened_malloc.so
+  fi
```
pastalian: Why not install both unconditionally?
As far as I can tell, there's no clear reason not to.

dev-libs/hardened_malloc/hardened_malloc-13.ebuild
```diff
@@ -0,0 +1,29 @@
+# Copyright 2025 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+DESCRIPTION="Hardened allocator designed for modern systems."
+HOMEPAGE="https://github.com/GrapheneOS/hardened_malloc"
+SRC_URI="https://github.com/GrapheneOS/hardened_malloc/archive/refs/tags/${PV}.tar.gz"
```
pastalian: Please rename it to include the package name.
```suggestion
SRC_URI="https://github.com/GrapheneOS/hardened_malloc/archive/refs/tags/${PV}.tar.gz -> ${P}.tar.gz"
```

tmak2002: i updated the Ebuild and Manifest

stkw0:
Thanks. It has been merged, you can close this PR

For the future, if you happen to do another PR and you need to merge it, you can use app-portage/pram. This would add the "Closes" tag, needed to automatically close the PR once it's merged.

PR #347, opened 2025-07-12
Why are there links to non official ollama repo's?

DevenBL:
removed SRC_URI's that pointed to [negril's ](https://github.com/negril/gentoo-overlay-vendored/) personal repository. It just looks real suspicious and does not need to be there.


mgorny:
This removes the `${P}-vendor.tar.xz` archive that is necessary to build the package. Did you even test this change?

PR #348, opened 2025-07-18
www-plugins/firefoxpwa: add 2.15.0

filips123:


stkw0:
Please, fix the next pkgcheck warning:  `TooManyCrates: version 2.15.0: CRATES variable has 416 crates, consider using crates tarball instead`

filips123:
Can you please still merge this PR, and I'll fix that in the next release? I will likely be easier for me if I just update pycargoebuild and my ebuild template to have it generated properly in the next release.

demize:
There’s a lot of information there about creating it manually, but in most circumstances you should be fine with just the one that `pycargoebuild` creates. Upload it somewhere (if you’re the upstream, adding it to each release is *perfect* for GURU) and add it to SRC_URI. 

You can take a look at how I do it in https://github.com/gentoo/guru/blob/master/dev-vcs/jj/jj-0.31.0.ebuild if you want a fairly simple example, though jj doesn’t have any git dependencies to worry about.

PR #351, opened 2025-07-22
app-misc/watchexec: use dependency files; version bump

ingenarel:
fixes the following pkgcheck warnings:

app-misc/watchexec
  TooManyCrates: version 2.1.2: CRATES variable has 440 crates, consider using crates tarball instead
  TooManyCrates: version 2.2.1: CRATES variable has 443 crates, consider using crates tarball instead
  TooManyCrates: version 2.3.0: CRATES variable has 487 crates, consider using crates tarball instead

also adds 2.3.2, 9999 

@lo48576 pinging you since you're the maintainer


lo48576:
Thank you for pinging me.

Just a note:
- Checked (using `diff -Naur`) that `vendor/` in the tarballs are identical to the files generated by  `cargo vendor` of corresponding version.
- Ebuild files looks good to me.

Personaly I'm a bit concerned about someone being required to create and host vendored crates tarball... since, generally speaking, a third-party files (hosted and/or generated) can be a security risk (as [recent incident at Arch AUR repo](https://lists.archlinux.org/archives/list/aur-general@lists.archlinux.org/thread/7EZTJXLIAQLARQNTMEW2HBWZYE626IFJ/) for example) and it is more bothersome for downstream users to audit tarballs than ebuilds and patches.
However, other than that, I have no reasons to ignore [TooManyCrates](https://pkgcore.github.io/pkgcheck/man/pkgcheck.html#toomanycrates) lint, so I think this Pull Request is OK.

<small>(It might be good if trusted CI/CD or something can be set up, but I have no idea how good and convenient it is because I've never used GitHub CI.)</small>

ingenarel:
> Personally I'm a bit concerned about someone being required to create and host vendored crates tarball... since, generally speaking, a third-party files (hosted and/or generated) can be a security risk (as [recent incident at Arch AUR repo](https://lists.archlinux.org/archives/list/aur-general@lists.archlinux.org/thread/7EZTJXLIAQLARQNTMEW2HBWZYE626IFJ/) for example) and it is more bothersome for downstream users to audit tarballs than ebuilds and patches.

i don't have full knowledge on how dependency files like this are handled, but i think the build should fail if someone modified the vendor files because hash mismatch or something (again, i don't have that much knowledge about it so take this with a bowl of salt)

> It might be good if trusted CI/CD or something can be set up, but I have no idea how good and convenient it is because I've never used GitHub CI.

about the ci/cd, i actually have my own script to create vendor files easily for rust and go projects (haven't added node yet) https://github.com/ingenarel/gentoo-dev-scripts/blob/master/crdeps

will probably even make it a package for guru after i make some good docs for it and improve the logic a little bit more.

closing this pr and gonna merge and push it in a few minutes to dev branch because you said that the pr looks okay enough

Schievel1:
> i don't have full knowledge on how dependency files like this are handled, but i think the build should fail if someone modified the vendor files because hash mismatch or something (again, i don't have that much knowledge about it so take this with a bowl of salt)


Yes exactly, the build fails when the hash of the tarball does not match with what the digest says. 
But what if the one who made the tarball and the digest manipulated it? 
For ::gentoo we can trust the devs, but for ::guru the onboarding process isn't very rigorous.

PR #352, opened 2025-07-23
gui-libs/gtk4-layer-shell: add support for ~x86

aruhier:
gtk4-layer-shell is a dependency of gui-apps/swaync-0.12, which supports `~x86`. Therefore, add the support to this dependency.

@pHlt7, as you noted this in [your comment](https://github.com/gentoo/guru/commit/7d81327474ea1afe1d9353112a1c152cecb38d9c#r162677442), I guess that you have a x86 installation to test this. I cannot test it myself, can you confirm to me that it works?

Thanks

### DIFF
--- a/gui-libs/gtk4-layer-shell/gtk4-layer-shell-1.1.1-r1.ebuild
+++ b/gui-libs/gtk4-layer-shell/gtk4-layer-shell-1.1.1-r1.ebuild
@@ -9,7 +9,7 @@ SRC_URI="https://github.com/wmww/${PN}/archive/refs/tags/v${PV}.tar.gz -> ${P}.t
 
 LICENSE="MIT"
 SLOT="0"
-KEYWORDS="~amd64"
+KEYWORDS="~amd64 ~x86"
 
 IUSE="examples doc test smoke-tests introspection vala"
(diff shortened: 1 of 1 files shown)


aruhier:
I don't know if x86 is being deprecated, I just realized that @pHlt7 had to drop the x86 support for swaync, and recommended here to deal with gtk4-layer-shell: https://github.com/gentoo/guru/commit/7d81327474ea1afe1d9353112a1c152cecb38d9c#r162677442

I'm happy to any advice, and if we should drop the explicit x86 support or not.

stkw0:
I am not sure about if it's deprecated. What I noticed recently is that sometimes ebuilds are added with support for x86 by default, probably because it was common to do it that way. Regardless about if it's deprecated or not, I think it's better to only add the architecture for which it has been tested  (which probably is amd64 or maybe some arm) and afterward, if someone needs x86, to keyword them. Some distros are already deprecating x86, so I will rather prefer to drop x86 if a dependency don't support it, rather than keywording all dependencies if there is no use case for it.

aruhier:
Sounds good! Then I'll close this PR. If anyone needs it, feel free to add the keyword.

PR #358, opened 2025-08-08
games-roguelike/cataclysm-dda: add clang use

donjuanplatinum:
Signed-off-by: Donjuanplatinum <donjuan@donplat.top>


pastalian:
Could you explain in the commit message why this change is needed? Does the clang build fail without CLANG=1 flag?

PR #359, opened 2025-08-13
dev-util/mx: new package, add 9999

IlitriumNumerus:
Add support for the mx compiler tool used by GraalVM. I plan to create the ebuild for GraalVM as well, so this PR is the first step. 
This is my first ebuild, I would appreciate if you could give me some advice and tips !


IlitriumNumerus:
Thanks for your feedback !
I replaced the live-ebuilds by the latest tags versions (for major version 6 and 7) and I used the python-single-r1 eclass (distutils not fitting this package). Could you review the new ebuilds ?
I've also opened an [access request](https://bugs.gentoo.org/961829).

dev-util/mx/mx-6.53.2.ebuild
```diff
@@ -0,0 +1,25 @@
...
+SLOT="0"
+RDEPEND="${PYTHON_DEPS}"
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
+KEYWORDS="~amd64"
+
+src_install() {
+	SITE_PKG="$(python_get_sitedir)/${PN}"
+	python_moduleinto ${SITE_PKG}
+	python_domodule *
+	chmod 0755 "${D}/${SITE_PKG}/mx" || die "Failed to chmod mx executable"
```
eli-schwartz: The upstream code isn't well set up for "proper" installation, but it's absolutely not sensible to install the *entire* git repository into `sitedir/mx/`.

This also then results in `sitedir/mx/mx` being a bash script that tries to run `sitedir/mx/mx.py` using several different versions of python, but doesn't respect Gentoo's EPYTHON which means that PYTHON_DEPS` is accomplishing nothing. :(

I would recommend instead installing src/mx as a module, and writing your own simple python script to install to /usr/bin that simply imports the module directly.

Please also figure out which data files actually need to be installed. It may make sense to install them to /usr/share/mx and patch the application itself to find the resources there instead of relative to `__file__`.

eli-schwartz (changes requested):

dev-util/mx/mx-6.53.2.ebuild
IlitriumNumerus: Took me quite some time to figure out this mess (the way modules were done prior version 7 were really messed up and version 7 also had to deal with backwards compatibility) but now it should be all good.

PR #361, opened 2025-08-19
dev-games/godot-dotnet: new package, add 4.4.1-r1

dsafxP:
This package is a fork of the `dev-games/godot::gentoo` package with .NET support.

Originally, I wanted to submit .NET support for Godot to the `dev-games/godot::gentoo` package as an IUSE flag, but upon contacting the current maintainer as of writing this (`ionen@g.o`) I was told that no maintainer was willing to pick up .NET support for it (possibly due to the increased complexity of the script)

Given this situation, I made it into a separate package with .NET support already included, since this is a fork I made this PR to clear any concerns regarding it and gain feedback.

The package hard-blocks `dev-games/godot::gentoo`, there's no point in maintaining complex coexistence for minimal benefit. Having two installations of Godot (one with .NET support and one without) is pointless as Godot with .NET support can do everything the regular version can do and it may be confusing for end-users.

I hope it is evident that .NET support is critical to the workflow of many developers (including myself).

https://bugs.gentoo.org/941089


dev-games/godot-dotnet/godot-dotnet-4.4.1-r1.ebuild
```diff
@@ -0,0 +1,509 @@
...
+xunit.runner.visualstudio@2.4.5
+"
+
+inherit desktop dotnet-pkg python-any-r1 flag-o-matic scons-utils
+inherit shell-completion toolchain-funcs xdg
+DESCRIPTION="Multi-platform 2D and 3D game engine with a feature-rich editor"
+HOMEPAGE="https://godotengine.org/"
+SRC_URI="
+	https://downloads.tuxfamily.org/godotengine/${PV}/godot-${PV}-stable.tar.xz
+	https://github.com/godotengine/godot/releases/download/${PV}-stable/godot-${PV}-stable.tar.xz
```
stkw0, line 260: What is the difference between both? They seem to be both a release of godot stable¿?

dsafxP:
Yes, I am aware. I'm opening the PR just in case anyone bats an eye due to it being a fork. I'll be closing it soon.

dev-games/godot-dotnet/godot-dotnet-4.4.1-r1.ebuild
dsafxP: I'm unsure, since that's what was set in the original script.

But, yes, you seem to be correct.

PR #362, opened 2025-08-19
app-shells/sheldon: new package, add 0.8.5

ursm:


stkw0:
Thanks por the PR! It's missing the signed-off-by tag but the ebuild looks good to me.

We don't currently merge PR for new packages. While we appreciate receiving PR as a way to review the ebuilds before they are added, you should register at GURU as explained in https://wiki.gentoo.org/wiki/Project:GURU/Information_for_Contributors to actually add them to GURU.

ursm:
Thanks a lot!

PR #363, opened 2025-08-20
FOR REVIEW ONLY games-strategy/7k-ambition: new package, add 3.2.2 and 9999

Infiltrator:


games-strategy/7k-ambition/7k-ambition-3.2.2.ebuild
```diff
@@ -0,0 +1,64 @@
...
+	default
+	eautoreconf
+}
+
+src_configure() {
+	econf \
+		--prefix="${EPREFIX}/opt/games/${P}" \
+		--datadir="${EPREFIX}/opt/games/${P}/share" \
+		--sysconfdir="${EPREFIX}/opt/games/${P}/etc" \
+		--localedir="${EPREFIX}/opt/games/${P}/share/locale"
```
stkw0: Why using /opt/games instead of the standard paths?

games-strategy/7k-ambition/7k-ambition-3.2.2.ebuild
```diff
@@ -0,0 +1,64 @@
...
+LICENSE="GPL-2"
+if [[ ${PV} == *9999* ]]; then
+	KEYWORDS=""
+	SLOT="9999"
+else
+	SLOT="$(ver_cut 1-2)"
+	KEYWORDS="~amd64"
+fi
+
+DEPEND:="dev-libs/boost
```
stkw0: It should be. 
```
DEPEND="dev-libs/boost:=
 ...
```

Infiltrator: Ah.  And a version number after it too?  So can I do `dev-libs/boost:>=1.74`?

stkw0: In that case that would be >=dev-libs/boost-1.74:=
For a more proper explanation you can look at the devmanual: https://devmanual.gentoo.org/general-concepts/dependencies/index.html#slot-dependencies

games-strategy/7k-ambition/7k-ambition-3.2.2.ebuild
```diff
@@ -0,0 +1,64 @@
...
+else
+	SLOT="$(ver_cut 1-2)"
+	KEYWORDS="~amd64"
+fi
+
+DEPEND=">=dev-libs/boost-1.74:=
+	media-libs/libsdl2[X,video]:=
+	media-libs/openal:=
+	net-misc/curl:=
+	net-libs/enet:1.3:="
```
stkw0: It appears duplicated here. You were already depending on a slot.

games-strategy/7k-ambition/7k-ambition-3.2.2.ebuild
Infiltrator: Okay, I added `:=` to the libraries.

games-strategy/7k-ambition/7k-ambition-3.2.2.ebuild
Infiltrator: Okay, so I misunderstood https://devmanual.gentoo.org/general-concepts/dependencies/index.html#slot-dependencies.

Updated.

stkw0: You could have net-libs/enet:1.3= but not net-libs/enet:1.3:=

games-strategy/7k-ambition/7k-ambition-3.2.2.ebuild
stkw0: Is there no reason, it should follow the standard paths

Infiltrator: It was to try to make it so that multiple versions could be installed simultaneously.

It seems to work with the standard paths too, though.

Will it work now?  (Updated.)

Infiltrator: Nope, nevermind.  It doesn't work.

I don't have any way for now on how to separate packages to be installable in the same time.  If you could suggest some changes, I would take the advice.

stkw0: It's a bit more complicated to implement than that and for this case there is no much to gain

games-strategy/7k-ambition/7k-ambition-3.2.2.ebuild
```diff
@@ -0,0 +1,63 @@
...
+}
+
+src_compile() {
+	emake AR="$(tc-getAR)"
+}
+
+src_install() {
+	emake DESTDIR="${D}" install
+
+	insinto "/opt/games/${P}/share/pixmaps"
```
stkw0: you still install things to opt

games-strategy/7k-ambition/7k-ambition-3.2.2.ebuild
```diff
@@ -0,0 +1,63 @@
...
+	default
+	eautoreconf
+}
+
+src_configure() {
+    econf \
+        --prefix="${EPREFIX}/usr/${P}" \
+        --datadir="${EPREFIX}/usr/share/${P}" \
+        --sysconfdir="${EPREFIX}/etc/games/${P}" \
+        --localedir="${EPREFIX}/usr/share/${P}/locale"
```
stkw0: you can remove all the defined paths as the default ones would be enough

games-strategy/7k-ambition/7k-ambition-3.2.2.ebuild
Infiltrator: Okay, we will leave it for the time being.

games-strategy/7k-ambition/7k-ambition-3.2.2.ebuild
```diff
@@ -0,0 +1,43 @@
+# Copyright 1999-2025 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit autotools desktop toolchain-funcs xdg
```
Infiltrator, line 6: It's still not completely working because xdg doesn't set the desktop icon.

stkw0: If I remember you were installing the icons in /opt which is not the correct path to install icons and therefore it didn't find anything

Infiltrator: Yeah, we changed that and we're not doing anything in /opt anymore.

stkw0: The xdg eclass don't install desktop files or icons by itself, it's only responsible for reloading desktop, icon and/or mimeinfo cache info if they are installed in the correct path

Infiltrator: Okay, I think I got it working.  How's this now?

stkw0: I was trying to build it and seems it fails to build with sys-devel/gettext-0.23.2, with version 0.22.5-r2 it works. Please check if with version 0.23.1-r1 it works and if not, add a `BDEPEND="<=sys-devel/gettext-0.23.0"`.

Also, you can remove pkg_postinst and pkg_postrm as it's redundant. For the live ebuild, you don't need to specify the KEYWORDS, so you could do something like 
```
DESCRIPTION="Seven Kingdoms: Ambition"
HOMEPAGE="https://seven-kingdoms-ambition.sourceforge.io/"

if [[ ${PV} == *9999* ]]; then
	inherit git-r3
	EGIT_REPO_URI="https://git.code.sf.net/p/seven-kingdoms-ambition/code"
	EGIT_BRANCH="next"
else
	SRC_URI="https://github.com/Infiltrator/${PN}/archive/refs/tags/v${PV}.tar.gz -> ${P}.tar.gz"
	KEYWORDS="~amd64"
fi

SRC_URI+=" https://sourceforge.net/projects/seven-kingdoms-ambition/files/${PN}.png/download -> ${PN}.png"
```

About the icons, I was checking and the png that is downloaded is not really an icon, so it goes to a different path and therefore the icon cache is not updated by xdg (only the desktop one).

Infiltrator: Yeah, the gettext error is a bit strange, since it implies that anything except 0.20 would not work, but it works with 0.22.5-r2.  Anyway, I tested with 0.23.1-r1, it didn't work, and I added the line.

Removed pkg_post*.

Changed the KEYWORDS.

What do you mean that the png is not really an icon?  What are the requirements for it to be an icon, and so end up in the proper path?

stkw0: Looking at https://specifications.freedesktop.org/icon-theme-spec it seems it's assumed that an icon must be a square, having the same width and height., which is not the case for that image.

stkw0: btw about the image, I don't think it's wrong, I am just explaining why it didn't get installed as an icon.

gettext seems to be an utility used for building the program, so it's a BDEPEND (not a DEPEND or a RDEPEND).

Infiltrator: Ah, my bad.  I didn't read what you said carefully enough.

Having said that, gettext gets compiled into the binary, but not dynamically linked.  So it does not need to go into RDEPEND, right?  Does it still need to be in DEPEND, since its headers are included?

stkw0: I didn't checked, but you are right. If it's used for compiling the application, it needs to be in DEPEND. If it links or dynamically loads (through dlopen) or uses any resource it provides at runtime, then it needs to be in RDEPEND also. In BDEPEND is usually for tools like compilers and so

I was looking at the project now and I see they have different options that you can enable/disable. The two interesting ones are the --disable-nls and --disable-multiplayer. Adding something like this should work
```
IUSE="nls multiplayer"
DEPEND="
        nls? ( <=sys-devel/gettext-0.23.0 )
        multiplayer? ( 
                   net-libs/enet:1.3=
                   net-misc/curl
       )
       media-libs/libsdl2[X,video]
       media-libs/openal
"
.....
src_configure() {
    local myeconfargs=(
        $(use_enable nls)
        $(use_enable multiplayer)
    )
    econf "${myeconfargs[@]}"
}
```
Then you should update metadata.xml with the description of the multiplayer USE or else pkgcheck will complain.
I didn't tested but seems if multiplayer is disabled it also turns off the dependency on curl.

stkw0: I think you changed the make_desktop_entry to not use an absolute path and seems it's against the specs:
```
 * As described in the Icon Theme Specification, icon file extensions are not
 * allowed in .desktop files if the value is not an absolute path.
```

The filesize of the PNG file seems to be different than what is recorded in the Manifest file:
```
!!! Fetched file: 7k-ambition.png VERIFY FAILED!
!!! Reason: Filesize does not match recorded size
!!! Got:      203178
!!! Expected: 197382
```
This puzzled me. If upstream change the image it would break the build, so if it was not because due to an issue on your side, it may be better to not specify an icon, or upload it to a different place.

In metadata.xml, you don't need to add a description for the `nls` USE as it's a global use.

Also, I just realize a mistake on my side. It should be `<sys-devel/gettext-0.23.0` and not `<=`

(I am sorry it has taken that long, but we are almost there :D also I am sorry for my mistakes, I am not an expert. I ended up learning some things by reviewing this ebuild)

Infiltrator: Ah, I see.  I was following https://devmanual.gentoo.org/eclass-reference/desktop.eclass/, which says:

> the icon to use in the menu entry
> this can be relative (to /usr/share/pixmaps) or a full path to an icon

If that's not right, I will change it back.

The manifest is my bad.  I updated the icon to be square, but forgot to update the manifest.  I'll do that now.  No, apparently I won't do it now.  :smiling_face_with_tear: 

So, I tried `ebuild 7k-ambition-3.2.2.ebuild manifest`, which just puts the same size and checksum back.  Then, I tried `emerge 7k-ambition` to see whether it would fetch the new file, but it seems to be using its cached version.  Then, I removed the line from the manifest and ran emerge again, hoping it would update something, but generating the manifest again gives the old values.  How do I update it properly?

Yeah, I was wondering why pkgcheck was complaining about multiplayer but not nls.  That makes sense.

I am actually wondering whether we should set the gettext version constraint to `0.22.5-r2` or lower specifically, since it's possible that a later 0.22 might break it.  The message does imply that any non-0.20 version should break it, but strangely it works for 0.22.5 and 0.21-12 (the version I have on my system).  Frankly, I don't understand it at all.

Oooh, or maybe I do understand!  I think that the `autopoint` and `gettext` versions have to match each other.  That's the case on my system.  And `autopoint --version` in Gentoo does give 0.22.5.  So I think we just have to say in the ebuild that the autopoint and gettext versions have to match each other.

No, nevermind.  On Gentoo they're both part of the gettext package, so their versions should always match.  I am back to not understanding it.  :D

> (I am sorry it has taken that long, but we are almost there :D also I am sorry for my mistakes, I am not an expert. I ended up learning some things by reviewing this ebuild)

Oooh, I didn't realise that.  From what I understood, you were the or one of the main guys for guru.  Thank you for all the help so far!

(You don't happen to be a maintainer in any other distros, do you?)

stkw0: > Then, I removed the line from the manifest and ran emerge again, hoping it would update something, but generating the manifest again gives the old values. How do I update it properly?

If you run `portageq distdir ` it will give you the path where all the dsitfiles are stored. If you remove the png from there and from Manifest file and then run ebuild manifest again it will fix it.

> I am actually wondering whether we should set the gettext version constraint to `0.22.5-r2` or lower

Yes, that would be safer. 

> Oooh, I didn't realise that. From what I understood, you were the or one of the main guys for guru. Thank you for all the help so far!

Nope, I am not a dev or anything. However, since Gentoo lacks manpower, I try to be of help whenever I can. I think that's the good part of being community-maintained, when a person is away, another can step in and help and so on.

And no, I don't collaborate in other distros, except filling bugs at times (:

stkw0 (approved):
I think it's good enough

PR #365, opened 2025-08-27
net-im/nheko: add 0.12.1

stkw0:


dev-libs/mtxclient/mtxclient-0.10.1.ebuild
CyberTailor, line 1: > `# Copyright 2020-2024 Gentoo Authors`

-2025

net-im/nheko/nheko-0.12.1.ebuild
```diff
@@ -0,0 +1,123 @@
...
+COMMON_DEPEND="
+	app-text/cmark:=
+	>=dev-cpp/blurhash-0.2.0:=
+	dev-cpp/cpp-httplib:=
+	dev-db/lmdb:=
+	dev-libs/libfmt:=
+	>=dev-libs/kdsingleapplication-1.1.0
+	>=dev-libs/mtxclient-0.10.1:=
+	dev-libs/olm
+	>=dev-libs/qtkeychain-0.14.1-r1:=
```
CyberTailor: A better way to write this dependency is `dev-libs/qtkeychain:=[qt6(+)]`

net-im/nheko/nheko-0.12.1.ebuild
```diff
@@ -0,0 +1,123 @@
+# Copyright 2020-2024 Gentoo Authors
```
CyberTailor: -2025

net-im/nheko/nheko-0.12.1.ebuild
stkw0: Will change it. Also, I think that gst-plugins- packages can be unrestricted as the version on the tree is already higher than what is required

PR #366, opened 2025-09-03
FOR REVIEW: games-strategy/7k-ambition: require gcc-12 for multiplayer

Infiltrator:
I don't know the best way to go about this.

Due to something as yet unknown, compiling with gcc 14 causes a desync in multiplayer, so we need to build with gcc-12 (gcc-13 is unknown and untested).

So we need to check that we are building with gcc-12 and, if not, tell the user to this package to use gcc-12.

### DIFF
--- a/games-strategy/7k-ambition/7k-ambition-3.2.4.ebuild
+++ b/games-strategy/7k-ambition/7k-ambition-3.2.4.ebuild
@@ -26,6 +26,7 @@ SLOT="0"
 IUSE="+nls +multiplayer +music"
 RESTRICT="music? ( bindist ) mirror"
 
+BDEPEND="multiplayer? ( sys-devel/gcc:12 )"
 DEPEND="
 	dev-libs/boost:=
 	nls? ( <=sys-devel/gettext-0.22.5-r2 )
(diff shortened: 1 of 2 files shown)


Infiltrator:
> Which optimizations flags do you use?

`-fsigned-char -mfpmath=387 -ffloat-store -g -O2 -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -fcf-protection -Wdate-time -D_FORTIFY_SOURCE=2 -D_REENTRANT`


> Can you check with gcc-15?

Good idea.  I will do that soon and let you know.


> Is some related upstream bug about it?

I posted about it just now.  We'll see what 3dfxdude says about it: https://7kfans.com/forums/viewtopic.php?t=1579

Ideally, he would just switch over to gcc 14 and do another release.  However, from my experience, this will either not happen or take a very long time.

Infiltrator:
Ah, yes, I forget about that.  Not sure why it's not public.  Creating an account is free, though.

I have created a bug on the tracker: https://sourceforge.net/p/skfans/bugs/5/

I do not know whether the bug will get any updates, though.  PRs on SourceForge take a very long time to get a response

games-strategy/7k-ambition/7k-ambition-3.2.4.ebuild
```diff
@@ -47,6 +47,15 @@ src_unpack() {
 }
 
 src_prepare() {
+	if use multiplayer ; then
+		gcc --version | grep 12\. || \
```
stkw0: btw you should use `if tc-is-gcc && ver_test $(gcc-version) -ne 12`  instead and add `BDEPEND="multiplayer? ( sys-devel/gcc:12 )"` and also made a new revision

Infiltrator:
I had a problem emerging gcc 15, so that is delayed.  Once I get that up, I will test with it.

Infiltrator:
Cool.  Thanks again for the help!

PR #367, opened 2025-09-04
FOR REVIEW: 7k ambition gcc precision

Infiltrator:
Not sure whether export is the correct way to do it in the ebuild.


stkw0:
Seems fine. Add a new revision for the non-live ebuild so it triggers a rebuild to users who could have merged the buggy ebuild

Infiltrator:
Cool.  So rename it to `7k-ambition-3.2.4-r1.ebuild`, right?  Do I need to rename the -9999 one too?  My understanding of that is that -9999 is just the rolling latest dev version, so I *think* that it's not needed here, right?

Infiltrator:
Ah.  And if I do copy it, do I fix it only in the -r1; or still fix it in both?

PR #368, opened 2025-09-05
mail-client/betterbird-bin: drop 128.13.0, add 140.2.1

BluCobalt:
The previous 128.13.0 ebuild was 404ing, so I took the opportunity to update it.


stkw0:
Thank you for the effort. Could you regenerate the Manifest files? It contains entries that are not the ones used for GURU

stkw0:
Yes. Thanks

PR #372, opened 2025-09-09
Feature/qttermtcp 0.79

bogogil:
QtTermTCP is a multi-platform version of BPQTermTCP. It supports TCP access to a bpq32 or linbpq node either over a local LAN or the Internet.


pastalian:
Qt5 support on Gentoo is deprecated and will eventually be removed.
I don't think adding a new package that only supports Qt5 is a good idea.
https://github.com/gentoo/gentoo/blob/e7e4cb3a6b72c4a918d8ab2bb04983d9ead406ca/profiles/package.deprecated#L20-L58

PR #373, opened 2025-09-19
gui-wm/river: fix man build in 0.3.12

xz-dev:
Fix: https://bugs.gentoo.org/962978


xz-dev:
> I am surprised the patch also changes /bin/sh to /usr/bin/sh as the later don't seems to exist in Gentoo. Did you made the patch yourself or it comes from upstream? If so would be nice to add the reference to it.

Thanks for the reminder, I update the patch.

PR #375, opened 2025-10-01
x11-themes/qogir-icon-theme: add 2025.02.15

joostruis:


stkw0:
Thanks. It's still missing the signoff tag.
Also would be better to rebase the commits into a single one

PR #399, opened 2025-11-21
app-misc/lf: add 38, 9999

ingenarel:
logic changes:
- use an ebuild template

@ephemer4l pinging you since you're the maintainer, if you give the green signal i'll push it to the dev branch


app-misc/lf/lf-38.ebuild
```diff
@@ -0,0 +1,79 @@
+# Copyright 1999-2025 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+inherit go-module shell-completion desktop xdg
```
antecrescent: Sort these alphabetically.

app-misc/lf/lf-38.ebuild
```diff
@@ -0,0 +1,79 @@
...
+
+LICENSE="MIT"
+
+# echo "# dependency licenses:"; printf 'LICENSES+=" '
+# go-licenses report ./... 2>/dev/null | awk -F ',' '{ print $NF }' | sort --unique | tr '\n' ' '; echo '"'
+
+# dependency licenses:
+LICENSES+=" Apache-2.0 BSD-3-Clause MIT "
+SLOT="0"
+IUSE="+static +man bash-completion zsh-completion fish-completion"
```
pastalian: These USE flags (man, shell completions) are not encouraged.
https://projects.gentoo.org/qa/policy-guide/installed-files.html#pg0301
https://projects.gentoo.org/qa/policy-guide/installed-files.html#pg0305

ingenarel: yp, i had a discussion about it in the irc cz i thought that bash-completion and zsh-completion is on the global index so i should use them

but later holgerh, ztrawhcse, and negril explained to me that they were leftovers before the policy change

i'll fix the changes soon in a few moments

* leaving this comment as info for future contributors

PR #403, opened 2025-12-01
dev-util/tokei: add 13.0.0, 9999

ingenarel:
features:

- specify a minimum rust version
- use an ebuild template

@lo48576 pinging you since you're the maintainer, if you give me the green signal i'll merge it to dev


dev-util/tokei/tokei-13.0.0.ebuild
```diff
@@ -0,0 +1,262 @@
+# Copyright 2017-2026 Gentoo Authors
```
gonsoos: Nit: It's not 2026 yet. Same or the live ebuild.

lo48576: Oops, for some reason I used `2017-2026` in 12.1.2 ebuild. It's my bad.

ingenarel: fixed it (i felt something was off with the years but didn't realize the year was wrong), if there are no more problems, give a thumbs up and i'll merge it to dev

dev-util/tokei/tokei-9999.ebuild
```diff
@@ -0,0 +1,57 @@
...
+	}
+else
+	SRC_URI="https://github.com/XAMPPRocky/tokei/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+	SRC_URI+=" ${CARGO_CRATE_URIS}"
+	KEYWORDS="~amd64"
+	RESTRICT="mirror"
+fi
+
+LICENSE="|| ( Apache-2.0 MIT )"
+
```
lo48576: I'm preoccupied with this empty line between `LICENSE`s, but it's harmless so no problem :-)

lo48576 (approved):

dev-util/tokei/tokei-9999.ebuild
ingenarel: okay i'll fix that, and i'll need to fix a few other problems too such as src_unpack

PR #440, opened 2026-02-15
www-plugins/tridactyl-native: new package, add 0.5.0

ingenarel:
i talked about this in the #gentoo-dev-help a bit however didn't got any response when i was online, hence i'm currently keeping this as pr and if everything is okay i can merge this to dev with pram

i'll paste the irc logs in here:

```
2026-02-14 21:15:02	ingenrelNeoJesus	so i'm trying to package some nim projects that i use, and i tried to start with tridactyl's native messenger, however i tried to make an ebuild with the nimble eclass that's in ::guru and it didn't work, then i tried to use CyberTailor 's nimbus to compile the project in a sepearete dir and then i saw that for some reason it wasn't compiling and nimbus was just throwing errors, i decided
2026-02-14 21:15:04	ingenrelNeoJesus	to ditch the nimble eclass alltogether, because from what i've seen it was using `nimbus` under the hood, and i made this: https://gist.githubusercontent.com/ingenarel/8390794ff433ea1b5998760509ee398a/raw
2026-02-14 21:20:18	ingenrelNeoJesus	i messed around with nimble for a while (i am not a nim dev and this is my first time even touching a project related to nim), and even tho from what i've seen nimble doesn't offer first class dependency tarball support like you see with go/cargo, i was able to find a workaround
2026-02-14 21:25:52	ingenrelNeoJesus	the depfile generation process was shown in the ebuild paste i linked, however ig i still should explain it, it uses find and file to get a list of all the files inside the vendor dir and their mimetypes, then it only filters files that are "application/json" or "text/*", then it pipes the filelist to tar and generates a dependency tarball
2026-02-14 21:27:46	ingenrelNeoJesus	i should again say that i just basicially took an educated guess and generated a file this way, the json manifest may not be needed, altho i haven't tested it out yet, however when i was writing it it felt the right way to parse what was neccesary and ignoring the binary files that also get generated
2026-02-14 21:28:08	ingenrelNeoJesus	s/manifest/manifests
2026-02-14 21:43:08	ingenrelNeoJesus	now i have a few questions, first of all can i write an ebuild like this or do i absolutely need to write an eclass myself or try to fix/work with the nimble eclass? second of all i still have doubts on how i did the src_compile function, should i move the dependency dir inside the ${WORKDIR}/${P} in the src_configure stage? third of all if CyberTailor or any other person that has
2026-02-14 21:43:10	ingenrelNeoJesus	experience with nim or working with nim downstream, was the nimble eclass the way to go and i'm just renventing the wheel?
2026-02-14 21:43:48	ingenrelNeoJesus	i also just read https://github.com/gentoo/gentoo/blob/b825c4280dc809870568b6280f787ac82105f4bb/eclass/golang-build.eclass#L64 so i guess i can keep the src_compile like how it is now?
```

if this is okay to merge or if i need to change stuff please let me know


www-plugins/tridactyl-native/tridactyl-native-0.5.0.ebuild
```diff
@@ -0,0 +1,47 @@
+# Copyright 1999-2026 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
```
gonsoos: 9, since you're not depending on incompatible eclasses.

www-plugins/tridactyl-native/tridactyl-native-0.5.0.ebuild
```diff
@@ -0,0 +1,47 @@
+# Copyright 1999-2026 Gentoo Authors
```
gonsoos: This is a new package created in 2026, so just `2026`.

www-plugins/tridactyl-native/tridactyl-native-0.5.0.ebuild
```diff
@@ -0,0 +1,47 @@
...
+		--nimbleDir:"${WORKDIR}/nimbledeps"\
+		--useSystemNim\
+		build
+}
+
+src_install(){
+	exeinto /usr/libexec/tridactyl
+	doexe native_main
+	sed -i -e "s|REPLACE_ME_WITH_SED|/usr/libexec/tridactyl/native_main|" ./tridactyl.json
+	local target_dirs=( /usr/lib{,64}/mozilla/native-messaging-hosts )
```
gonsoos: `usr/lib64` looks incorrect. I think that should be `usr/$(get_libdir)`, like in https://gitweb.gentoo.org/repo/gentoo.git/tree/gnome-extra/gnome-browser-connector/gnome-browser-connector-42.1-r1.ebuild#n35.

But I suggest you get another opinion on that on IRC.

www-plugins/tridactyl-native/tridactyl-native-0.5.0.ebuild
```diff
@@ -0,0 +1,47 @@
...
+		--localDeps\
+		--nimbleDir:"${WORKDIR}/nimbledeps"\
+		--useSystemNim\
+		build
+}
+
+src_install(){
+	exeinto /usr/libexec/tridactyl
+	doexe native_main
+	sed -i -e "s|REPLACE_ME_WITH_SED|/usr/libexec/tridactyl/native_main|" ./tridactyl.json
```
gonsoos:
```suggestion
	sed -i -e "s|REPLACE_ME_WITH_SED|${EPREFIX}/usr/libexec/tridactyl/native_main|" ./tridactyl.json
```

Missing `|| die`.

www-plugins/tridactyl-native/metadata.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "https://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+  <maintainer type="project">
```
gonsoos:
```suggestion
  <maintainer type="person">
```

www-plugins/tridactyl-native/tridactyl-native-0.5.0.ebuild
```diff
@@ -0,0 +1,47 @@
...
+S="${WORKDIR}/native_messenger-${PV}"
+
+LICENSE="BSD-2"
+
+SLOT="0"
+
+KEYWORDS="~amd64"
+
+src_compile() {
+	nimble\
```
gonsoos, line 37: Missing `BDEPEND` for that.

www-plugins/tridactyl-native/tridactyl-native-0.5.0.ebuild
```diff
@@ -0,0 +1,47 @@
...
+		--verbose\
+		--offline\
+		--localDeps\
+		--nimbleDir:"${WORKDIR}/nimbledeps"\
+		--useSystemNim\
+		build
```
gonsoos: Missing `|| die`.

www-plugins/tridactyl-native/tridactyl-native-0.5.0.ebuild
```diff
@@ -0,0 +1,47 @@
...
+}
+
+src_install(){
```
gonsoos: The missing space in `){` is inconsistent with `src_prepare`. Please fix that.

www-plugins/tridactyl-native/tridactyl-native-0.5.0.ebuild
```diff
@@ -0,0 +1,47 @@
...
+	exeinto /usr/libexec/tridactyl
+	doexe native_main
+	sed -i -e "s|REPLACE_ME_WITH_SED|/usr/libexec/tridactyl/native_main|" ./tridactyl.json
+	local target_dirs=( /usr/lib{,64}/mozilla/native-messaging-hosts )
+	for target_dir in "${target_dirs[@]}"; do
```
gonsoos, line 54: `target_dir` is not local, but should be.

gonsoos (changes requested):

www-plugins/tridactyl-native/tridactyl-native-0.5.0.ebuild
ingenarel: wait i actually didn't know EAPI 9 is officially a thing now, i guess i'll try to read about it when i'm free

www-plugins/tridactyl-native/tridactyl-native-0.5.0.ebuild
ingenarel: isn't the `target_dir` gonna be be local to that loop anyway because of variable scope?

gonsoos: No, non-local variable state gets preserved for later functions.
https://projects.gentoo.org/pms/9/pms.html#the-state-of-variables-between-functions

www-plugins/tridactyl-native/tridactyl-native-0.5.0.ebuild
gonsoos: https://projects.gentoo.org/pms/9/eapi-cheatsheet.pdf

www-plugins/tridactyl-native/tridactyl-native-0.5.0.ebuild
ingenarel: i did a bit of testing on this one myself, on an amd64 system, `get_libdir` returns `lib64`, however when tested with librewolf-bin from the [librewolf](https://codeberg.org/librewolf/gentoo.git) overlay the tridactyl extension isn't able to detect the native messenger, so i'm currently keeping both lib and lib64 install

gonsoos: Could be the same situation as with firefox-bin, which expects plugins to reside in`usr/lib`. Explained in the commit message  https://github.com/gentoo/gentoo/commit/3d78ce7f7d4173f0cd57b958a5702ac5a63e44ec.

> on an amd64 system `get_libdir` returns `lib64`

Not necessarily, as it depends on your profile/ABI. On profiles with x32 ABI, it returns `libx32` for example.
That's why I proposed to keep the `usr/lib` install, but replace `usr/lib64` with `usr/$(get_libdir)`.

ingenarel: ah i thought you wanted me to do only the `usr/$(get_libdir)` install, i'll fix the ebuild

PR #454, opened 2026-03-30
net-irc/halloy: add missing libxkbcommon runtime dependency

afiw0:
Halloy dlopens libxkbcommon at runtime, the dependency was probably not noticed because most systems have libxkbcommon installed anyway.
Revbumped because the dependency is a runtime one.


agireud (approved):
Thank you!

PR #455, opened 2026-03-31
Updates to dev-util/xrt ebuild

istitov:
- Correcting build with boost >= 1.89.0 (otherwise fails at configure)
- Adding ms-gl into dependencies (build fails otherwise)
- see #456 as well


shuber2:
Stumbled over the same missing ms-gl dep here. Thanks.

PR #460, opened 2026-04-12
sci-ml/ollama: add 0.20.5

pva:
Version bump.
Go version increased since https://github.com/ollama/ollama/blob/v0.20.5/go.mod#L3


AshyPinguin:
The commit does not include the ` Signed-off-by: Username <username@example.com>`

This is needed because of [regulation 2](https://wiki.gentoo.org/wiki/Project:GURU#The_regulations) (I think).

PR #461, opened 2026-04-17
x11-misc/pywal16: add 3.8.14

superlemon57:
added 2026 version

### DIFF
--- a/x11-misc/pywal16/pywal16-3.8.14.ebuild
+++ b/x11-misc/pywal16/pywal16-3.8.14.ebuild
@@ -1,0 +1,29 @@
+# Copyright 2026 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+PYTHON_COMPAT=( python3_{12..13} )
+DISTUTILS_USE_PEP517=setuptools
+
+inherit distutils-r1 optfeature
+
+DESCRIPTION="16 colors fork of pywal"
+HOMEPAGE="https://github.com/eylles/pywal16"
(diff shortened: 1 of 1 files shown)


stkw0:
Seems the Manifest file was not updated. Also remember to use the standard commit message format (or use `pkgdev commit --signoff` which would do it for you)

x11-misc/pywal16/pywal16-3.8.14.ebuild
```diff
@@ -0,0 +1,29 @@
+# Copyright 2026 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+PYTHON_COMPAT=( python3_{12..13} )
```
stkw0, line 6: py 3.14 can be enabled

x11-misc/pywal16/pywal16-3.8.14.ebuild
```diff
@@ -0,0 +1,29 @@
...
+
+DESCRIPTION="16 colors fork of pywal"
+HOMEPAGE="https://github.com/eylles/pywal16"
+SRC_URI="https://github.com/eylles/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~arm64"
+
+RDEPEND="media-gfx/imagemagick[jpeg]"
```
stkw0, line 19: If unit tests depends on this, which they likely do, I think this dependency should be also a DEPEND or BDEPEND when `test` is enabled

PR #467, opened 2026-05-02
x11-themes/nordzy-cursors: add 2.5.0

SammyCyber:
Adds new version 2.5.0 and changes SRC_URI to gitlab, which is the new source of the package. Also uses the tar.bz2 file instead of tar.gz as `pkgcheck scan` recommends it over the latter.

`pkgcheck scan` also returns `MissingRemoteId: missing <remote-id type="gitlab">gboehm/Nordzy-cursors</remote-id> (inferred from URI 'https://gitlab.com/gboehm/Nordzy-cursors')`. I think the way to fix this is to modify `metadata.xml`, but I'm not sure how exactly to (do we want to add the gitlab link and keep the github one, as it's a mirror, or replace it entirely?).

I was also thinking of adding two local use flags, xcursors and hyprcursors, to limit the type of cursor that is installed. This would avoid the user having to install a huge number of files, but I'm not sure if this would go against the guidelines in CONTRIBUTING.md


stkw0:
Thanks for the PR.

Yes, you should add the metadata of the new remote-id without removing the Github one.
Once the older version is removed the Github remote-id should be removed too. 

About the USE flag, I would leave it as it's now. @MrRoy is the maintainer of this package, so unless he is in favorable to that I would not change that part.

SammyCyber:
Thanks for the info. I've added the new remote-id and also changed the bugs-to link as the github one doesn't exist anymore.

PR #476, opened 2026-05-18
gui-apps/quickshell: fix missing USE constraint i3? ( X )

nado:
Quickshell.I3 qml module wasnt installed without enabling x11 dependency, the cmake file itself doesn't enforce that dependency but will silently skip i3 module without x11, see https://github.com/AvengeMedia/DankMaterialShell/issues/2437

Quickshell with X enabled also didnt compile without X enabled on qtbase.

I took the liberty to remove a space character to reduce (by 1 line) the diff between `0.3.0` and `9999`

I wasn't sure whether I should've opened the PR against master or dev.

### DIFF
--- a/gui-apps/quickshell/quickshell-0.3.0.ebuild
+++ b/gui-apps/quickshell/quickshell-0.3.0.ebuild
@@ -36,10 +36,11 @@ REQUIRED_USE="
 	toplevel-management? ( wayland )
 	hyprland?            ( wayland )
 	screencopy?          ( wayland )
+	i3? ( X )
 "
 
 RDEPEND="
-	dev-qt/qtbase:6=[dbus,vulkan]
+	dev-qt/qtbase:6=[dbus,vulkan,X?]
 	dev-qt/qtsvg:6=
 	dev-qt/qtdeclarative:6=
 	x11-libs/libdrm
--- a/gui-apps/quickshell/quickshell-9999.ebuild
+++ b/gui-apps/quickshell/quickshell-9999.ebuild
@@ -33,10 +33,11 @@ REQUIRED_USE="
 	toplevel-management? ( wayland )
 	hyprland?            ( wayland )
 	screencopy?          ( wayland )
+	i3? ( X )
 "
 
 RDEPEND="
-	dev-qt/qtbase:6=[dbus,vulkan]
+	dev-qt/qtbase:6=[dbus,vulkan,X?]
 	dev-qt/qtsvg:6=
 	dev-qt/qtdeclarative:6=
 	jemalloc? ( dev-libs/jemalloc )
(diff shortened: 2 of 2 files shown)


stkw0:
Do you want to handle this @ceres-sees-all (as a maintainer of this package)?

AshyPinguin:
I think it doesn't make it depend on `X` is because it also enables sway things

As it says in the ebuild:
https://github.com/gentoo/guru/blob/3fe57979da1c567771c46b2ec1aef2cc58ac929b/gui-apps/quickshell/quickshell-0.3.0.ebuild#L82

It also says so in the metadata.xml:
https://github.com/gentoo/guru/blob/3fe57979da1c567771c46b2ec1aef2cc58ac929b/gui-apps/quickshell/metadata.xml#L28

nado:
If you inspect the sources CMakeLists.txt from the successive src directories:

https://github.com/quickshell-mirror/quickshell/blob/59e9c47b0eb48a9e4bcf9631fa062ee939bd2e83/src/CMakeLists.txt#L28-L30

`src/CMakeLists.txt`
```cmake
if (X11)
	add_subdirectory(x11)
endif()
```

https://github.com/quickshell-mirror/quickshell/blob/59e9c47b0eb48a9e4bcf9631fa062ee939bd2e83/src/x11/CMakeLists.txt#L14-L16

`src/x11/CMakeLists.txt`
```cmake
if(I3)
	add_subdirectory(i3)
endif()
```

And i3 module is built exclusively in this directory `src/x11/i3`

ceres-sees-all:
Sorry for the wait, github seems to not be sending me emails :/

Annoying that this is how upstream decided to do it, but i fully agree with your logic.
Pushed in https://github.com/gentoo/guru/commit/37b1ba75a824eb502b25921eef34d8daeb66a837

AshyPinguin:
> Sorry for the wait, github seems to not be sending me emails :/
> 
> Annoying that this is how upstream decided to do it, but i fully agree with your logic.
> Pushed in https://github.com/gentoo/guru/commit/37b1ba75a824eb502b25921eef34d8daeb66a837

Btw this (i3) also enables sway stuff which doesn't depend on X. So it will probably be smarter to leave it how it was imo.

nado:
> > Sorry for the wait, github seems to not be sending me emails :/
> > Annoying that this is how upstream decided to do it, but i fully agree with your logic.
> > Pushed in [37b1ba7](https://github.com/gentoo/guru/commit/37b1ba75a824eb502b25921eef34d8daeb66a837)
> 
> Btw this (i3) also enables sway stuff which doesn't depend on X. So it will probably be smarter to leave it how it was imo.

I couldn't find any other ref to I3 or I3_IPC outside of x11 subdirectories cmake file, could you point me to what you're referring to? If you're right, best solution would be to add einfo lines at pkg_postinst I guess.

ceres-sees-all:
> > Sorry for the wait, github seems to not be sending me emails :/
> > Annoying that this is how upstream decided to do it, but i fully agree with your logic.
> > Pushed in [37b1ba7](https://github.com/gentoo/guru/commit/37b1ba75a824eb502b25921eef34d8daeb66a837)
> 
> Btw this (i3) also enables sway stuff which doesn't depend on X. So it will probably be smarter to leave it how it was imo.

I disagree, currently, if the user has has I3 enabled but not X, they will not build support for  I3, so are missing functionality they expect. With this change, if they only want sway they will have more functionality than needed, which while annoying is vastly preferable to the other option. I dont really see how this could be fixed without pretty heavily patching the CMakeLists

PR #479, opened 2026-05-23
sci-electronics/ghdl-yosys-plugin: new package

Endergon:


sci-electronics/ghdl-yosys-plugin/ghdl-yosys-plugin-6.0.0.ebuild
```diff
@@ -0,0 +1,39 @@
...
+
+DEPEND="
+	sci-electronics/yosys
+	=sci-electronics/ghdl-6.0.0*
+"
+RDEPEND="${DEPEND}"
+
+src_compile() {
+	local gcc_ver=$(gcc-major-version)
+	local adalib_path="/usr/lib/gcc/${CHOST}/${gcc_ver}/adalib"
```
stkw0, line 26: This means the ebuild depends on gcc[ada], right?

Endergon: No, this ebuild does not depend on gcc[ada], but ghdl itself does.

PR #481, opened 2026-05-27
app-admin/createrepo-rs: new package, version 0.1.8

jamesarch:
Pure Rust RPM repository metadata generator — dnf/yum-compatible, zero FFI, single static binary.

- GitHub: https://github.com/jamesarch/createrepo_rs
- crates.io: https://crates.io/crates/createrepo_rs
- License: GPL-2.0-or-later
- Uses cargo.eclass for build


thesamesam (changes requested):
Unfortunately the PR has several invalid changes and it looks likely to be LLM generated.

mgorny:
Closing given that this ebuild could have never worked, and therefore it clearly wasn't tested prior to submission.

PR #484, opened 2026-06-02
app-arch/brzip: new package

matijaskala:
Brotli is great for compressing documentation because it decompresses much faster than bzip2 while keeping similar compression ratio. Unfortunately the raw brotli format doesn't contain magic bytes by which it could be recognized. Brzip encapsulates brotli-compressed data into the [brv3](https://github.com/madler/brotli/blob/master/br-format-v3.txt) frame format.


app-arch/brzip/brzip-0.3.2.ebuild
```diff
@@ -0,0 +1,21 @@
+# Copyright 1999-2026 Gentoo Authors
```
ceres-sees-all: As this is a new package, it can just be "Copyright 2026 Gentoo Authors"

app-arch/brzip/brzip-0.3.2.ebuild
```diff
@@ -0,0 +1,21 @@
...
+inherit meson
+
+DESCRIPTION="A compression utility based on Brotli algorithm"
+HOMEPAGE="https://gitlab.com/ms1888/brzip"
+SRC_URI="https://gitlab.com/ms1888/brzip/-/archive/${PV}/${P}.tar.gz"
+
+LICENSE="GPL-3+"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
```
ceres-sees-all: Can omit this line, as empty IUSE is default behaviour

ceres-sees-all:
Apart from a few little details looks all good! For GURU, pull requests are usually only for review, so you can apply for access following the instructions [on the wiki](https://wiki.gentoo.org/wiki/Project:GURU/Information_for_Contributors) and push yourself.

PR #485, opened 2026-06-06
app-misc/tradicted-trading-journal: new package 1.0.0

tradicted:
New package: tradicted-trading-journal 1.0.0

A free open-source offline-first desktop trading journal for Linux.
Features trade log, analytics dashboard, journal with calendar view,
multiple portfolio support and a trading plan checklist.
All data stored locally, no account required.

Homepage: https://www.tradicted.com
License: MIT


app-misc/tradicted-trading-journal/tradicted-trading-journal-1.0.0.ebuild
```diff
@@ -0,0 +1,32 @@
...
+}
+
+src_install() {
+	insinto /opt/${PN}
+	newins "${P}.AppImage" "${PN}.AppImage"
+	fperms 0755 /opt/${PN}/${PN}.AppImage
+
+	make_wrapper ${PN} /opt/${PN}/${PN}.AppImage
+
+	domenu "${FILESDIR}/${PN}.desktop"
```
mgorny, line 31: Any reason not to use `make_desktop_entry`?

app-misc/tradicted-trading-journal/tradicted-trading-journal-1.0.0.ebuild
```diff
@@ -0,0 +1,32 @@
+# Copyright 2026 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+DESCRIPTION="Free open-source desktop trading journal by Tradicted"
+HOMEPAGE="https://www.tradicted.com"
+SRC_URI="https://github.com/tradicted/tradicted-journal/releases/download/v${PV}/tradicted-trading-journal-${PV}.AppImage -> ${P}.AppImage"
```
mgorny, line 8: Please build from source. We don't want binary crap with ton of bundled dependencies.

mgorny (changes requested):
Plea

eli-schwartz (changes requested):
This is:
- brand new software 
- zero actual users
- you don't appear to actually be a Gentoo user, you've just submitted it to a bunch of distros / package managers, for self promotion


It also fails to generally take into account the rules and regulations for contributing to this project, which is unsurprising given point 3, and also... unlikely to change.

I don't see a significant reason to go above and beyond the call of duty and also try to actually review the content of the PR -- what I've already described is already enough to say this can't be merged.

PR #486, opened 2026-06-07
README.md: make it explicit PR should not be used to add new packages

stkw0:
Adds a note making it clear PR for new packages are not welcomed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ The [GURU regulations](https://wiki.gentoo.org/wiki/Project:GURU#The_regulations
 
 *See [wiki.gentoo.org/wiki/Project:GURU](https://wiki.gentoo.org/wiki/Project:GURU) for more information on the project.*
 
+Note: By policy, PR that adds a new package are NOT accepted.
+
 </td>
 <td width="27%" style="border-style:solid; border-radius:10px;">
 


README.md
```diff
@@ -19,6 +19,8 @@ The [GURU regulations](https://wiki.gentoo.org/wiki/Project:GURU#The_regulations
 
 *See [wiki.gentoo.org/wiki/Project:GURU](https://wiki.gentoo.org/wiki/Project:GURU) for more information on the project.*
 
+Note: By policy, PR that adds a new package are NOT accepted.
```
Leo3418, line 22: Where is the text of the policy?  Could you add a link to it?

mgorny:
While I don't mind making such a policy explicit (though it should be a consensus within GURU), I'm slightly worried that we'd just see more people requesting access to dump stuff and forget about it.

stkw0:
>  I'm slightly worried that we'd just see more people requesting access to dump stuff and forget about it.

To filter that, we could slightly modify how access is requested asking users to include their motivation so then people can explain if they noticed some package that requires more love, some popular software missing or self-promoting (which case we can deny access).

mgorny:
Per the new policy, we now need to accept new packages via PRs.

PR #493, opened 2026-06-13
fix desktop entry referencing non-existent /usr/bin/Waterfox

debiplobi:
This fixes app launchers not detecting waterfox-bin after installing it


stkw0:
The commit doesn't follow the standard commit message and lacks the required Certificate of origin (see https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin ). 

You can use `pkgdev commit --signoff` to assist you with it.

PR #504, opened 2026-07-03
sys-apps/pkgit:  add package manager

zofh:


sys-apps/pkgit/metadata.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "https://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+  <maintainer type="person">
+    <email>blod@duck.com</email>
+    <name>any</name>
+  </maintainer>
```
negril, line 7: This is not a know contributor. Nor would it be valid

sys-apps/pkgit/pkgit-9999.ebuild
```diff
@@ -0,0 +1,23 @@
+# Copyright 1999-2025 Gentoo Authors
```
negril: We have 2026

sys-apps/pkgit/pkgit-1.1.3.ebuild
```diff
@@ -0,0 +1,23 @@
...
+#upstream does not provide release tarballs/snapshots, using git-r3.
+inherit git-r3
+
+DESCRIPTION=""
+HOMEPAGE="https://git.symlinx.net/pkgit/about"
+LICENSE="GPL-2"
+SLOT="0"
+
+EGIT_REPO_URI="https://git.symlinx.net/pkgit"
+EGIT_COMMIT="1.1.3"
```
negril, line 15: Just no. Do not pin the version or use `EGIT_COMMIT` for a -9999.

Where did you get this idea from?

sys-apps/pkgit/pkgit-9999.ebuild
```diff
@@ -0,0 +1,23 @@
...
+HOMEPAGE="https://git.symlinx.net/pkgit/about"
+LICENSE="GPL-2"
+SLOT="0"
+
+EGIT_REPO_URI="https://git.symlinx.net/pkgit"
+EGIT_COMMIT="1.1.3"
+KEYWORDS=""
+
+RDEPEND="
+	dev-vcs/git
```
negril: You already inherit `git-r3`, which adds a depend on `dev-vcs/git`.

sys-apps/pkgit/pkgit-9999.ebuild
```diff
@@ -0,0 +1,23 @@
...
+	net-misc/curl
```
negril: `curl` does not show up in the source, where is this coming from?

sys-apps/pkgit/pkgit-9999.ebuild
```diff
@@ -0,0 +1,23 @@
+# Copyright 1999-2025 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+#upstream does not provide release tarballs/snapshots, using git-r3.
+inherit git-r3
+
+DESCRIPTION=""
```
negril: This is empty...

sys-apps/pkgit/pkgit-1.1.3.ebuild
```diff
@@ -0,0 +1,23 @@
+# Copyright 1999-2025 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+#upstream does not provide release tarballs/snapshots, using git-r3.
```
negril, line 6: We don't do -9999 only packages.

negril (changes requested):
The things noted in https://github.com/gentoo/guru/commit/fdd4a00f817e10dec642f99447a0eb2f313ab4f4
and the things still present in https://github.com/gentoo/guru/commit/185cd62ecf23bbed00270da20a296ae519198cdf apply here as well.

As a trusted contributor. NACK

negril:
Not sufficient.

sys-apps/pkgit/pkgit-1.1.3.ebuild
masterwolf-git: So what should we do in this situation, fork the repository and make releases? Or is there a better way?

falbrechtskirchinger: On GitHub (etc.), you can do what I did here:

https://github.com/gentoo/guru/blob/827b85ee1330217f118f77d16dcb31a969e79f5c/sys-devel/clang-bloomberg-p2996/clang-bloomberg-p2996-21.0.0_p20250702.ebuild#L19-L20

Otherwise, create and host a snapshot some other way.

masterwolf-git: But this way you cannot use keywords, right?
Thanks!

falbrechtskirchinger: You can absolutely use keywords here. A live ebuild is unkeyworded, for one, because it pulls from a moving target, making the build unrepeatable.

Snapshots are deterministic and are treated like any other release.

masterwolf-git: Oh, thank you very much for the explanation!

PR #508, opened 2026-07-13
added LLVM22 support for xwayland-satellite

ph0llux:
I've tested the appropriate build using LLVM 22 without any problems (LVM22 is available (and partially default) now for a lot of packages).


stkw0:
The newer version already supports llvm 22. This PR can be closed.

PR #510, opened 2026-07-16
dev-util/opencode-bin: Bump versions to 1.15.13, 1.16.2, 1.17.20, 1.18.1 and drop 1.1.48, 1.2.6, 1.15.12

red-avalanche:
# Description
Bump dev-util/opencode-bin to the latest upstream stable release (1.18.1) and add the latest patch version of each intermediate minor series between the previous GURU latest (1.15.12) and upstream latest.
## Versions added:
- 1.15.13 (latest 1.15.x, supersedes 1.15.12)
- 1.16.2 (latest 1.16.x)
- 1.17.20 (latest 1.17.x)
- 1.18.1 (latest upstream stable)
## Versions dropped:
- 1.1.48, 1.2.6, 1.15.12 (old snapshots / superseded)
# Other changes:
- Added ~arm64 keyword support. Upstream now ships opencode-linux-arm64.tar.gz alongside the existing x64 tarball; SRC_URI updated to use arch-conditional URIs.
- SRC_URI reformatted to wrap long lines (resolves ExcessiveLineLength pkgcheck warning, limit 120 chars).
Per GURU regulation #8 (maintainability), only the latest patch of each minor series is packaged rather than every tagged release (~26 between 1.15.12 and 1.18.1). This matches the existing repo pattern of one ebuild per minor series.
# QA Status
- pkgcheck scan --net dev-util/opencode-bin — clean (no warnings/errors)
- pkgcheck scan --commits HEAD~2..HEAD — clean
- Manifest regenerated via pkgdev manifest; 8 distfile entries (4 versions × 2 arches), all BLAKE2B/SHA512 verified
- ebuild ... install phase tested for all 4 versions (amd64); each:
- Passed manifest verification
- Unpacked successfully
- Installed to temporary image dir with correct layout (usr/bin/opencode)
- Binary executes and reports correct version string (opencode --version)
- arm64 distfiles verified present (HTTP 200) and tarball structure confirmed identical to amd64 (single opencode binary at root); not runtime-tested on arm64 hardware


eli-schwartz:
> * Added ~arm64 keyword support. Upstream now ships opencode-linux-arm64.tar.gz alongside the existing x64 tarball; SRC_URI updated to use arch-conditional URIs.

FYI, this should really have been done all along. If the source is arch-conditional, then it should be *described thusly* in `SRC_URI`, it doesn't matter if there is only one KEYWORDS.

For example, this makes it clearer to readers, what needs to be changed to enable additional arches. It also prevents users from installing the wrong architecture because they tried to use `ACCEPT_KEYWORDS=**` (which is also how you accept a live ebuild!)

The fix to SRC_URI could be done as a preparatory cleanup commit.

red-avalanche:
Thanks for the feedback. Is this a blocking change you're requesting, or just an FYI?

PR #514, opened 2026-07-18
Rust OpenStack client - separate

NiklausHofer:
I've packaged openstack_cli and openstack_tui from https://github.com/gtema/openstack. These are two separate programs developed in one repository.

Because of that I am not sure how best to package them - separate ebuilds or just one for both of them? So I've done both!

This MR is for the separate version. You can find the combined ebuild in [this other MR here](https://github.com/gentoo/guru/pull/513). Please only merge one of them...


masterwolf-git:
This is nice.

I understand the TUI needs the CLI?

If so, I think it could be a conditional dependency based on the TUI use flag.

NiklausHofer:
> I understand the TUI needs the CLI?

What makes you think that? I'm not saying you're wrong. It would explain a lot of things (why tui compiles so much faster, why there are no tests for tui, ...).

But I don't think it does. I tested just now by installing tui only. The osc is not present on my system. But despite that, tui works fine.

PR #515, opened 2026-07-19
sci-misc/llama-cpp: fix web UI

WGH-:
Since upstream PR https://github.com/ggml-org/llama.cpp/pull/22937, static assets are no longer vendored in the source code repository. Instead, they're now published on release to HuggingFace and (as of more recent llama.cpp releases) to GitHub release assets.

Update the ebuild to fetch the UI archive from GitHub, and unpack it to the correct directory.

This fixes nonfunctional web UI that returns 404 error instead of the page.

Note that this commit does not try to solve live 9999 ebuild: actually building the code with JavaScript tooling is much more difficult problem.

Closes: https://bugs.gentoo.org/979245


WGH-:
sci-misc/llama-cpp-0_pre9888 is no more, and new ebuilds, 0.3.0 and up, have similar fix incorporated.

PR #537, opened 2026-08-14
dev-python/ninja-syntax: new package, add 2026.04.06

IlitriumNumerus:
Add a simple python module, required to build GraalVM. It is linked to https://github.com/gentoo/guru/pull/359

---

Please check all the boxes that apply:

- [X] I have read the [GURU rules](https://wiki.gentoo.org/wiki/Project:GURU#Rules) and agree with them (note that they were last updated 2026-07-19).
- [X] I can submit this contribution in agreement with the [Copyright Policy](https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin).
- [X] I have certified the above via adding a `Signed-off-by` line to *every* commit in the pull request.
- [X] This contribution has not been created with the assistance of Natural Language Processing artificial intelligence tools, in accordance with the [AI policy](https://wiki.gentoo.org/wiki/Project:Council/AI_policy).
- [X] I have run `pkgcheck scan --commits --net` to check for issues with my commits.
- [X] I have included the build logs for all the packages I'm adding or modifying.
- [X] All e-mail addresses added to `metadata.xml` correspond to valid [Gentoo Bugzilla](https://bugs.gentoo.org/), and I will be reading the bug mail send to them, or none were added.


dev-python/ninja-syntax/ninja-syntax-20260406.ebuild
```diff
@@ -0,0 +1,28 @@
...
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+PYTHON_COMPAT=( python3_{11..15} )
+
+inherit python-single-r1
+
+DESCRIPTION="Python module for generating .ninja files"
+HOMEPAGE="https://github.com/ninja-build/ninja"
+SRC_URI="https://raw.githubusercontent.com/ninja-build/ninja/9cabe6a3f3e69ea7ef3665c97e46d4f14433f7d8/misc/ninja_syntax.py"
```
eli-schwartz: You can't do this, the file that is downloaded to $DESTDIR contains no version info.

dev-python/ninja-syntax/Manifest
```diff
@@ -0,0 +1 @@
+DIST ninja_syntax.py 8148 BLAKE2B aecdf3bbf3c852eb2ec04280ebffaeb73dade3c47717a05720eb208d54bdca2a6a0246a79b4ced75d2e83c047c51377417ab1bfbb84f621a9a2fc0b0fee4a35f SHA512 2ea1175b65d4b9473ea5cb1863397e5330f96776638d60a3d6ea17d067802fb7ceebafc1d1f51d16ea9939b1d5bd70a35c77360dd651ec1d9b6d45da0a543cc7
```
eli-schwartz: ... as seen here, the distfile will clash every time you update the package, and users will get checksum errors.

dev-python/ninja-syntax/ninja-syntax-20260406.ebuild
```diff
@@ -0,0 +1,28 @@
+# Copyright 2026 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+PYTHON_COMPAT=( python3_{11..15} )
```
eli-schwartz: Please note that Python 3.11 is now a no-op and should not be included in new ebuilds.

dev-python/ninja-syntax/ninja-syntax-20260406.ebuild
```diff
@@ -0,0 +1,28 @@
...
+
+inherit python-single-r1
```
eli-schwartz: Why use single-r1 out of curiosity? What if someone really wanted to install it for multiple python versions? For example if some other package wanted to depend on this package it would not be permitted for that package to, itself, use the "multiple python targets" eclass.

dev-python/ninja-syntax/ninja-syntax-20260406.ebuild
eli-schwartz: Also, have you really tested this on python 3.15? Impressive if so :) given that it's still a release candidate.

dev-python/ninja-syntax/ninja-syntax-20260406.ebuild
IlitriumNumerus: I could use python-r1 if it's necessary. I went for single-r1 in the first place, as the ebuild who will use it (mx) is single-r1. I don't know any other packages using ninja-syntax, but I guess I can switch to python-r1.

dev-python/ninja-syntax/ninja-syntax-20260406.ebuild
IlitriumNumerus: I did it to avoid downloading the whole repo just for a small file. Should I just use ninja versioned source code archives ? Another option would be to add a USE to gentoo's ninja ebuild.

dev-python/ninja-syntax/ninja-syntax-20260406.ebuild
IlitriumNumerus: Python 3.15 shouldn't break anything that is used in this module, so I assumed it would work on 3.15.

dev-python/ninja-syntax/ninja-syntax-20260406.ebuild
eli-schwartz: Regardless of whether you use the versioned source code archives or not, the downloaded filename merely has to include some distinctive factor.

Many ebuilds download `${PN}/archive/v${PV}.tar.gz` but since the default filename only has PV, and not PN, it would clash with any other package that has a similar version number. So SRC_URI supports "arrow renames".

PR #549, opened 2026-09-04
sci-ml/llmman: new package, add 0.1.0_p315

ericcurtin:
Adds [llmman](https://github.com/llmmanorg/llmman), a local model runner that serves the Ollama API (alongside OpenAI- and Anthropic-compatible ones) on port 17434, as `sci-ml/llmman` next to `sci-ml/ollama`.

- `sci-ml/llmman-0.1.0_p315`: upstream is a Rust crate whose `build.rs` also compiles a Go shim (`go-shim/`, `-buildmode=c-archive`), so the ebuild inherits both `cargo` and `go-module`. A single `src_unpack` calls `cargo_src_unpack` (which also extracts `${P}-deps.tar.xz` into `${WORKDIR}/go-mod`, the `GOMODCACHE` that `go-module.eclass` exports) and then `go-env_set_compile_environment`. Upstream tags are `bNNN` with `Cargo.toml` fixed at `0.1.0`, hence `0.1.0_p315` -> tag `b315`.
- `CRATES` (440 crates, all from crates.io) and the crate `LICENSE+=` were generated with `pycargoebuild 0.16.0` from the tagged `Cargo.lock`; Go module licenses (Apache-2.0, MIT, BSD, Unlicense) were collected with `go-licenses`, not `dev-go/lichen`, so please re-check.
- **Go dependency tarball is not hosted yet.** `SRC_URI` points at `https://github.com/llmmanorg/llmman/releases/download/b315/llmman-0.1.0_p315-deps.tar.xz`, following CONTRIBUTING's advice to have upstream release CI publish it. The `Manifest` therefore has DIST entries for the GitHub tarball and all crates but none for `-deps.tar.xz`; it can be generated with `GOMODCACHE="${PWD}"/go-mod go mod download -modcacherw` in `go-shim/` and `XZ_OPT='-T0 -9' tar -acf llmman-0.1.0_p315-deps.tar.xz go-mod` (about 400 MiB uncompressed) and the Manifest regenerated once hosted.
- `acct-group/llmman`, `acct-user/llmman` and the OpenRC/systemd service files mirror the ollama ones so `llmman serve` can run as a system service. `pkg_postinst` notes that llmman uses `llama-server` from `PATH` if present and otherwise downloads a prebuilt release into the service user's data dir on first use, since there is no llama.cpp package to depend on.
- No `/api/embed*` in llmman (embeddings only via `/v1/embeddings`); not relevant to packaging but mentioned for completeness.

**Testing:** On macOS (no Portage available): `cargo build --release` of the `b315` checkout with the Go shim built offline (`GOPROXY=off`) from a populated `GOMODCACHE` succeeded (332 crates compiled, `llmman --version` -> `llmman 0.1.0 (0e7a3ed)`); `go mod download`/`go mod verify` with `GOPROXY=off` against that cache passed ("all modules verified"); `bash -n` on the ebuilds and init script, `xmllint --noout` on the metadata files; Manifest DIST sizes/BLAKE2B/SHA512 computed from the downloaded distfiles.

Not verified: no `ebuild`/`emerge` run, no `pkgcheck scan --commits --net`, no build log; the deps tarball URL does not exist yet (see above).

> AI-assisted, reviewed before submitting.

---

Please check all the boxes that apply:

- [ ] I have read the [GURU rules](https://wiki.gentoo.org/wiki/Project:GURU#Rules) and agree with them (note that they were last updated 2026-07-19).
- [x] I can submit this contribution in agreement with the [Copyright Policy](https://www.gentoo.org/glep/glep-0076.html#certificate-of-origin).
- [x] I have certified the above via adding a `Signed-off-by` line to *every* commit in the pull request.
- [ ] This contribution has not been created with the assistance of Natural Language Processing artificial intelligence tools, in accordance with the [AI policy](https://wiki.gentoo.org/wiki/Project:Council/AI_policy).
- [ ] I have run `pkgcheck scan --commits --net` to check for issues with my commits.
- [ ] I have included the build logs for all the packages I'm adding or modifying.
- [ ] All e-mail addresses added to `metadata.xml` correspond to valid [Gentoo Bugzilla](https://bugs.gentoo.org/), and I will be reading the bug mail send to them, or none were added.


negril:
I've just glanced this for now. But given: 
> Models are pulled as OCI artifacts from any registry or directly from Hugging Face and are served by llama.cpp (llama-server) or vLLM.

Is there any merit to adding users and groups for every "local runner of the hour"? Gut feeling is for the majority of people it wouldn't make a difference and we should stick to one `user`:`group` pair if we can.
Read, use `ollama`:`ollama` here. And long term we should discuss having a single user for all the ggml/hugginface using stuff. Either via renaming the `ollama` user or coming up with a new one.

https://github.com/gentoo/guru/pull/549/changes?w=1#diff-aa3e53cbefb42fd7581987207e65225ab8d168da46e1cc9bbd889a31d1fcec14R506-R507
> GGUF models are run with llama-server, taken from PATH if present, otherwise a prebuilt release is downloaded into the llmman user's data directory on first use.

This seems flawed? `sci-misc/llama-cpp::guru` does exist and so this should depend on it.

More later.

ericcurtin:
Closing this, reads like a lack of interest

negril:
No it's called a review.

mgorny:
Violation of multiple rules, banning from Gentoo org.

